PR #9, opened 2017-05-31
Add a new method that keeps the original record after extraction

richwhitjr:
This new method keeps around the original record after doing the feature extraction so you get back a tuple2 of (Record, Features).

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
+++ b/core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
@@ -88,14 +88,18 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
   @transient private val dt: CollectionType[M] = implicitly[CollectionType[M]]
   import dt.Ops._
 
-  @transient private lazy val as: M[ARRAY] = input.map(fs.unsafeGet)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -123,11 +127,20 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
...
+  def featureValues[F: FeatureBuilder : ClassTag]: M[F] =
+    labeledFeatureValues.map(_._2)
+
+  /**
+   * Values of the extracted features, in the same order as names in [[featureNames]] with the
+   * original raw feature object.
+   * @tparam F output data type, e.g. `Array[Float]`, `Array[Double]`, `DenseVector[Float]`,
+   *           `DenseVector[Double]`
+   */
+  def labeledFeatureValues[F: FeatureBuilder : ClassTag]: M[(T,F)] = {
```
nevillelyh: "Labeled" is a little misleading? I thought it meant feature labels/names without reading the code. Any better suggestion?

richwhitjr: Yeah I was thinking about that.  I thought of Keyed also but that is also misleading especially with the distributed libs. Maybe recordFeatureValues?

PR #16, opened 2017-06-27
Make non serializable error explicit instead of "silent" NPE

ravwojdyla:
If fs is not serializable it would be null out by CC, which would
result in NPE, instead make sure that fs is not null out, and fail
earlier and with serialization exception.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
+++ b/core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
@@ -38,7 +38,10 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
   @transient private val dt: CollectionType[M] = implicitly[CollectionType[M]]
   import dt.Ops._
 
-  @transient private lazy val as: M[(T, ARRAY)] = input.map(o => (o, fs.unsafeGet(o)))
+  @transient private lazy val as: M[(T, ARRAY)] = {
+    val g = fs // defeat closure
+    input.map(o => (o, g.unsafeGet(o)))
+  }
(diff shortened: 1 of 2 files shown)


scio/src/test/scala/com/spotify/featran/scio/ScioTest.scala
```diff
@@ -31,14 +31,14 @@ class ScioTest extends PipelineSpec {
         .required(_._1)(OneHotEncoder("one_hot"))
         .required(_._2.toDouble)(MinMaxScaler("min_max"))
         .extract(sc.parallelize(data))
-      f.featureNames should containSingleValue (Seq(
+      f.featureNames should containSingleValue(Seq(
```
nevillelyh: keep this and the next one unchanged?

core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -38,15 +38,16 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
   @transient private val dt: CollectionType[M] = implicitly[CollectionType[M]]
   import dt.Ops._
 
-  @transient private lazy val as: M[(T, ARRAY)] = input.map(o => (o, fs.unsafeGet(o)))
+  private def as(@transient input: M[T], fs: FeatureSet[T]): M[(T, ARRAY)] =
+    input.map(o => (o, fs.unsafeGet(o)))
```
nevillelyh: Will this be executed twice, once in `aggregate` and once in `featureValuesWithOriginal`? Could be potentially expensive if some feature getter does expensive computation?

PR #17, opened 2017-07-10
Support multiple Specs and separate on output

richwhitjr:
The minor part of this PR is to add a function that allows the combination of Feature Specs easily.  The larger part though is adding the ability to extract multiple specs at the same time and separate out the resulting features by spec.  

The reason to do this is that common practice in libraries like TF is to have features in multiple columns.  You may want different labels or maybe one with weights.  It is also hard/slow to index/slice into existing arrays.  With this change you can pass in a List of specs and get by a list of features per row.  You can then write that data to disk however you like.  This avoid having to find out the size of each feature and indexing into the vectors to split out different label types.


nevillelyh:
LGTM overall but it does adds a bit of complexity.
What's the reason not to just use a large `FeatureSpec`, but prefix transformer names to indicate subsets, e.g. `fs1_id, fs1_vec, ... fs2_id, fs2_vec, ...`?

richwhitjr:
The issue is that you actually may want to write out multiple vectors in different fields.  Say a vector for users, and items plus 2 or 3 labels.  Without something like this you would have to grab the feature names and index into the vector correctly.  This is pretty easy in things like numpy but a nightmare in tensorflow.  Easier to only select the vectors you want on read in at train time.

nevillelyh:
Sorry out for conference right now haven't got around to this. Will look again and merge next week.

richwhitjr:
No rush, let me know if you have any ideas on how to simplify this new API.  One thought i had was to mimic the api of FeatureExtraction in the new MultiFeatureExtraction instead of subclassing as I do here.  So you can do below without having to use the special multi prefix functions.

```scala
val vectors: Seq[Seq[SparseVector[Double]]] = MultiFeatureSpec.specs(f, f2).extract(xs)
```

nevillelyh:
Agree, seems better to mimic `FeatureExtractor` in `MultiFeatureExtractor` instead of extending. The single extraction methods don't make sense in `Multi*` anyway.

Also pushed some minor changes to [neville/multi](https://github.com/spotify/featran/tree/neville/multi) branch. Got rid of the `FeatureBuilderContructor`s with a serialization hack. Take a look?

PR #18, opened 2017-07-11
VectorIdentity and NHotWeightedEncoder Transformers

richwhitjr:
The Vector Identity Transformer makes it easy to take a dense vector and push it through into the features.  NHotWeightedEncoder is similar to NHotEncoder but where each field has a weight.  For example when you have some probability that a label belongs on an item.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/NHotWeightedEncoder.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/NHotWeightedEncoder.scala
@@ -11,7 +11,7 @@ import com.twitter.algebird.Aggregator
  * @param name Name of the field
  * @param value Weight on the field
  */
-case class WeightedValues(name: String, value: Double)
+case class WeightedValue(name: String, value: Double)
 
 object NHotWeightedEncoder {
   /**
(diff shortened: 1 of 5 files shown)


core/src/main/scala/com/spotify/featran/transformers/NHotWeightedEncoder.scala
```diff
@@ -0,0 +1,52 @@
...
+import com.spotify.featran.FeatureBuilder
+import com.twitter.algebird.Aggregator
+
+/**
+ * Class for a weighted vector.  Also can be thought as
+ * a named sparse vector
+ * @param name Name of the field
+ * @param value Weight on the field
+ */
+case class WeightedValues(name: String, value: Double)
```
nevillelyh: This should be called `WeightedValue`? It's only 1 value.

core/src/main/scala/com/spotify/featran/transformers/NHotWeightedEncoder.scala
```diff
@@ -0,0 +1,52 @@
...
+  override val aggregator: Aggregator[Seq[WeightedValues], Set[String], Array[String]] =
+    Aggregators.from[Seq[WeightedValues]](_.map(_.name).toSet).to(_.toArray.sorted)
+
+  override def featureDimension(c: Array[String]): Int = c.length
+
+  override def featureNames(c: Array[String]): Seq[String] = c.map(name + "_" + _).toSeq
+
+  override def buildFeatures(a: Option[Seq[WeightedValues]], c: Array[String],
+                             fb: FeatureBuilder[_]): Unit = {
+    val as = a.map(_.map(n => (n.name, n.value)).toMap).getOrElse(Map.empty)
```
nevillelyh: `.map(...).toMap` is expensive.
Do this instead?
```scala
val as: Map[String, Double] = a match {
  case Some(xs) => xs.map(x => (x.name, x.value))(scala.collection.breakOut)
  case None => Map.empty
}
```

core/src/main/scala/com/spotify/featran/transformers/VectorIdentity.scala
```diff
@@ -0,0 +1,32 @@
...
+  override def featureDimension(c: Unit): Int = length
+
+  override def featureNames(c: Unit): Seq[String] = names
+
+  override def buildFeatures(a: Option[Seq[Double]], c: Unit, fb: FeatureBuilder[_]): Unit =
+    if(a.isDefined) fb.add(names.toIterator, a.get.toArray) else fb.skip(length)
+
+  override def encodeAggregator(c: Option[Unit]): Option[String] = c.map(_ => "")
+  override def decodeAggregator(s: Option[String]): Option[Unit] = s.map(_ => ())
+}
```
nevillelyh: You should `override def params` too to encode `length`

core/src/test/scala/com/spotify/featran/transformers/NHotWeightedEncoderSpec.scala
```diff
@@ -0,0 +1,19 @@
...
+
+import org.scalacheck.{Arbitrary, Gen, Prop}
+
+class NHotWeightedEncoderSpec extends TransformerProp("NHotWeightedEncoder") {
+  private implicit val labelArb = Arbitrary {
+    Gen.choose(1, 10).flatMap(Gen.listOfN(_, Gen.alphaStr))
+  }
+
+  property("default") = Prop.forAll { xs: List[List[String]] =>
+    val vecs = xs.map(_.map(n => WeightedValues(n, 0.5)))
```
nevillelyh: Better randomize weights here?

Issue #20, opened 2017-07-24
one hot encoder for large N


slhansen:
The one hot encoder takes a long time to run when the number of tokens is large. I believe the problem lies in the fact that the current implementation iterates over each element in the dictionary:

https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala#L45

I tried writing an implementation that uses a hash map instead of a list to store the tokens (plus their index), but it is still necessary to "inflate" the feature with the empty elements.

https://github.com/slhansen/featran/blob/master/core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala#L57-#L68

Since we can assume that only one element will be added it would be preferable to have the default be an empty feature vector and just add one element to it given the index.

slhansen:
@nevillelyh ^

richwhitjr:
Makes sense, in the skip function you can already specify N.  So we could be smarter about overriding this in the implementation functions as to not loop over the empty elements. 

So we could find the index into the one hot encoding list, skip up that, fill in the zero, then skip to the end.

slhansen:
@richwhitjr I tried the approach of making use of the the skip function n-1 times (see second link in original comment) but it is still slow. My use case is for encoding user IDs so n is ~10^7.

richwhitjr:
I think the issue here is that the subclasses don't implement skip(n) in an efficient way.  It may make sense to implement it in each subclass as needed for performance.  Right now with the default implementation I believe it is doing O(n) still even with the skipping.

https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/FeatureBuilder.scala#L43

nevillelyh:
Does either of you wanna submit a PR? Otherwise we can look into it.

PR #21, opened 2017-07-26
overwrite skip n function

slhansen:
Addresses part of the problem detailed in:

https://github.com/spotify/featran/issues/20

It is impractical to iteratively call skip n times for large n.

@richwhitjr @nevillelyh


nevillelyh:
You still need to fix `OneHotEncoder` to call `skip(n)` instead of looping over the list right?

slhansen:
@nevillelyh Plan was to do the one-hot-encoder rewrite in a separate PR

PR #22, opened 2017-07-27
use hash map instead of list for one hot encoder

slhansen:
Rewrite of one hot encoding that uses a hash map instead of a list. Needed for when the number of tokens is large. Addresses the issue detailed in:

https://github.com/spotify/featran/issues/20

@richwhitjr @nevillelyh

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
@@ -22,6 +22,8 @@ import java.net.{URLDecoder, URLEncoder}
 import com.spotify.featran.FeatureBuilder
 import com.twitter.algebird.Aggregator
 
+import scala.collection.mutable.HashMap
+
 object OneHotEncoder {
   /**
    * Transform a collection of categorical features to binary columns, with at most a single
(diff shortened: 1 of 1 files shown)


core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
```diff
@@ -31,20 +33,60 @@ object OneHotEncoder {
...
-  override def buildFeatures(a: Option[String], c: Array[String], fb: FeatureBuilder[_]): Unit =
-    c.foreach(s => if (a.contains(s)) fb.add(name + '_' + s, 1.0) else fb.skip())
-  override def encodeAggregator(c: Option[Array[String]]): Option[String] =
-    c.map(_.map(URLEncoder.encode(_, "UTF-8")).mkString(","))
-  override def decodeAggregator(s: Option[String]): Option[Array[String]] =
-    s.map(_.split(",").map(URLDecoder.decode(_, "UTF-8")))
+  extends Transformer[String, Set[String], HashMap[String, Int]](name) {
+
+  private def present(reduction: Set[String]): HashMap[String, Int] = {
+    val map = new HashMap[String, Int]
```
nevillelyh: You can use a map builder to avoid mutable map.

```scala
val b = Map.newBuilder[String, Int]
b += "a" -> 1
b.result
```

core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
```diff
@@ -31,20 +33,60 @@ object OneHotEncoder {
...
+
+    // if the set size is under 1000 first sort elements (also needed for unit tests)
+    val maxSizeForSorting = 1000
+    if (reduction.size <= maxSizeForSorting) {
+      reduction.toArray.sorted.zipWithIndex.foreach { case (k: String, i: Int) => map += k -> i }
```
nevillelyh: `zipWithIndex` creates a copy and is expensive. Use a `var i` and `while` loop over the array. For the `else` case, loop over `reduction.iterator`.

nevillelyh: Actually you can avoid this sorting if/else logic by using a `TreeMap` which sorts the keys.

core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
```diff
@@ -31,20 +33,60 @@ object OneHotEncoder {
...
+      reduction.zipWithIndex.foreach { case (k: String, i: Int) => map += k -> i }
+    }
+    map
+  }
+
+  override val aggregator: Aggregator[String, Set[String], HashMap[String, Int]] =
+    Aggregators.from[String](Set(_)).to(s => present(s))
+  override def featureDimension(c: HashMap[String, Int]): Int = c.size
+  override def featureNames(c: HashMap[String, Int]): Seq[String] =
+    c.toList.sortBy(_._2).map(name + '_' + _._1)
```
nevillelyh: You can eliminate the `.toList.sortBy` by using a `TreeMap`. Assuming the keys are always sorted, `_._2` should be continuously increasing as well?

core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
```diff
@@ -31,20 +33,60 @@ object OneHotEncoder {
...
+  override def buildFeatures(a: Option[String],
+                             c: HashMap[String, Int],
+                             fb: FeatureBuilder[_]): Unit = {
+    val n = c.size
+    a match {
```
nevillelyh: This may be more concise and readable.
```scala
val kv = for (k <-a; v <- c.get(k)) yield (k, v)
kv match {
  case Some((k, v)) => ...
  case None => fb.skip(n)
}
```

core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
```diff
@@ -31,20 +33,60 @@ object OneHotEncoder {
...
+            fb.skip(math.max(0, n-value-1))
+          }
+          case None => fb.skip(n)
+        }
+      }
+      case None => fb.skip(n)
+    }
+  }
+  override def encodeAggregator(c: Option[HashMap[String, Int]]): Option[String] =
+    c.map(_.toList.sortBy(_._2).map(e => URLEncoder.encode(e._1, "UTF-8")).mkString(","))
```
nevillelyh: See if this `.toList.sortBy` and the `zipWithIndex` below can be optimized?

core/src/main/scala/com/spotify/featran/transformers/OneHotEncoder.scala
slhansen: Could you elaborate on how to use `TreeMap`? The key is the name and the value is the position in the one hot encoded vector. In the rest of the class I always sort by the value (position). When adding entries to the `TreeMap` I wouldn't know their value until after all of the names have been added? I only added the sorting logic for the unit tests.

PR #37, opened 2017-08-18
support input rejection, fix #33

### LINKED ISSUE #33: Allow transformer to reject inputs
By adding a method to `FeatureBuilder`, so any `Transformer` in the spec can call it to reject an entire record.

nevillelyh:


core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -80,20 +80,19 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
...
 
   /**
-   * Values of the extracted features, in the same order as names in [[featureNames]] with the
-   * original input record.
+   * Values of the extracted features, in the same order as names in [[featureNames]] with
+   * rejections keyed on feature name and the original input record.
    * @tparam F output data type, e.g. `Array[Float]`, `Array[Double]`, `DenseVector[Float]`,
    *           `DenseVector[Double]`
    */
-  def featureValuesWithOriginal[F: FeatureBuilder : ClassTag]: M[(F, T)] = {
```
richwhitjr: Should we keep this function if you don't want to handle the rejections?

core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
```diff
@@ -20,17 +20,39 @@ package com.spotify.featran
...
 
+  private val _rejections: mutable.Map[String, FeatureRejection] = mutable.Map.empty
+
+  def reject(transformer: Transformer[_, _, _], reason: FeatureRejection): Unit = {
+    val name = transformer.name
+    require(!rejections.contains(name), s"Transformer $name already rejected")
+    _rejections(name) = reason
+  }
+
+  def rejections: Map[String, FeatureRejection] = {
```
richwhitjr: Should this be private also so that someone doesn't accidentally drop the rejections?

nevillelyh: It returns a copy so that shouldn't happen.
Plus we can't make it `private`, has to be `private[featran]` which kinda defeats the purpose.

PR #40, opened 2017-08-23
add java wrapper, fix #38

### LINKED ISSUE #38: Java wrapper?
We might wanna add a Java wrapper so that it could be used in a Java backend service.

nevillelyh:


java/src/main/java/com/spotify/featran/java/SerializableFunction.java
```diff
@@ -0,0 +1,24 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.java;
+
+import java.io.Serializable;
+
+public interface SerializableFunction<InputT, OutputT> extends Serializable {
```
richwhitjr: Not possible to do this with java 8 lambdas because of the serializable part?  Would be nice to use lambdas when specifying the function instead of creating an instance of this interface.  

```(InputT p) -> p.outputT```

nevillelyh: You can use lambda syntax at call site i.e. `.required(r -> r.getValue)` but `java.util.Function` is not serializable so this is a hack as argument type for `required`. It's actually copied as is from Beam.
Yes Java 8 is full of sadness. Everything (optional, stream, future) feels half done.

richwhitjr: That makes me said but sounds like the best approach in that case.

richwhitjr:
👍

PR #46, opened 2017-09-20
add java sparse array, fix #45

### LINKED ISSUE #45: support for sparse vectors in JFeatureExtractor
We would like to be able to extract features in Java as sparse vectors. Similar to the way you can do `extractor.featureValues[SparseVector[Float]]` in Scala.

nevillelyh:


java/src/main/scala/com/spotify/featran/java/JavaOps.scala
```diff
@@ -65,4 +65,71 @@ private object JavaOps {
...
+        indices = new Array[Int](n)
+        values = new Array[T](n)
+      }
+    }
+    override def add(name: String, value: Double): Unit = {
+      indices(i) = offset
+      values(i) = fp.fromDouble(value)
+      i += 1
+      offset += 1
+      if (indices.length == i) {
```
richwhitjr: Any reason not to use ArrayBuffer?  I believe the behavior you have coded here duplicates that class.   You can also give it a hintSize with the n from above.

nevillelyh: Duh I'll look into it.

nevillelyh: Actually no, since `ArrayBuffer` doesn't specialize for primitive and will use boxed `Integer`, `Float`, `Double`, etc., and I'll have to duplicate code to manually specialize them.

PR #49, opened 2017-10-03
add HeavyHitters, fix #48

### LINKED ISSUE #48: Frequency rank transformer
Feature request from an internal user:
"I want to transform a string feature to integer ids, where the most common string gets id 1, second most common gets id 2, etc. and have a cap at at 10000, so everything else gets mapped to 0. Is that easily doable with featran?
The original dataset might have say 1 billion strings (of which say 1 million are unique)."

Not sure if this can be done with approximation in a single pass. The naive approach would be building a `Map[String, Int]` in `reduce` phase and a priority queue before `map` phase.

nevillelyh:


andrewsmartin:
👍, only thing I'm not sure about is using `0.0` as the rank for items falling outside of the heavy hitters - could be misinterpreted as having the lowest rank and therefore highest frequency. Not sure if there's really a better choice though, maybe -1.0 or something? Anyway not a big concern.

andrewsmartin:
Ignore the above comment, I just saw the discussion in #48 :)

nevillelyh:
There's a serialization issue in `SketchMap`
https://github.com/twitter/algebird/pull/646

nevillelyh:
@maris382 please take a look

PR #55, opened 2017-10-16
port CombinatoricsUtils and remove commons-math3, fix #54

### LINKED ISSUE #54: Port commons-math3 CombinatoricsUtils logic over
So we can ditch the `commons-math3` dependency

nevillelyh:
DO NOT MERGE
Coverage broken


nevillelyh:
It's broken as in not 100% coverage ;)
It's mostly to avoid unnecessary dependencies, i.e. pulling in a huge jar for 1 function.

Issue #53, opened 2017-10-12
Make tests for probabilistic transformers black box


nevillelyh:
So that the test code doesn't depend on implementation details.
- Hash*Encoders (maybe)
- HeavyHitters
- QuantileDiscretizer

PR #56, opened 2017-10-17
Add NGramsTransformer, fix #52

### LINKED ISSUE #52: Add ngrams transformer
See https://github.com/tensorflow/transform/blob/master/tensorflow_transform/mappers.py#L399

andrewsmartin:
@nevillelyh PTAL. Gonna leave some comments / questions in a few places.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/NGrams.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/NGrams.scala
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.transformers
+
+import com.spotify.featran.FeatureBuilder
+
+import scala.collection.SortedMap
+
+/**
+ * Transform a collection of sentences, where each row is a `Seq[String]` of the words / tokens,
+ * into a collection containing all the n-grams that can be constructed from each row. The feature
+ * representation is an n-hot encoding (see [[NHotEncoder]]) constructed from an expanded vocabulary
+ * of all of the generated n-grams.
+ *
+ * N-grams are generated based on a specified range of `low` to `high` (inclusive) and are joined by
+ * the given `separator` (the default is " "). For example, an [[NGrams]] with
+ * `separator = ""`, `low = 2`, and `high = 3`, the row `["a", "b", "c", "d", "e"]` would produce
+ * the ngrams `["ab", "bc", "cd", "de", "abc", "bcd", "cde"]`.
+ *
+ * As with [[NHotEncoder]], missing values are transformed to [0.0, 0.0, ...].
+ */
+object NGrams {
+  /**
+   * Create a new [[NGrams]] instance.
+   *
+   * @param low the smallest size of the generated *-grams. Default is 1.
+   * @param high the largest size of the generated *-grams. Default is -1, which means n-grams will
+   *             be generated up to the full length of the `Seq[String]`
+   * @param sep a string separator used to join individual tokens. Default is " ".
+   */
+  def apply(name: String, low: Int = 1, high: Int = -1, sep: String = " ")
+  : Transformer[Seq[String], Set[String], SortedMap[String, Int]] = {
+    require(low > 0, "'low' must be a positive integer")
+    require(high > 0 || high == -1, "'high' must either be a positive integer, or -1")
+    new NGrams(name, low, high, sep)
+  }
+}
+
+private class NGrams(name: String, val low: Int, val high: Int, val sep: String)
+  extends NHotEncoder(name) {
+  override def prepare(a: Seq[String]): Set[String] = ngrams(a).toSet
+
+  override def buildFeatures(a: Option[Seq[String]],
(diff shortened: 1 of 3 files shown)


core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+  /**
+   * Create a new [[NGramsTransformer]] instance.
+   * @param separator a string separator used to join individual tokens
+   * @param low the smallest size of the generated *-grams
+   * @param high the largest size of the generated *-grams
+   */
+  def apply(name: String, separator: String, low: Int, high: Int): TfType =
+    new NGramsTransformer(name, separator, low, high)
+
+  def apply(name: String, high: Int): TfType = apply(name, "", 1, high)
```
andrewsmartin: Are these reasonable defaults to have, and are they even worth having?

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+                                val high: Int) extends NHotEncoder(name) {
+  override def prepare(a: Seq[String]): Set[String] = ngrams(a).toSet
+
+  override def buildFeatures(a: Option[Seq[String]],
+                             c: SortedMap[String, Int],
+                             fb: FeatureBuilder[_]): Unit =
+    super.buildFeatures(a.map(ngrams(_)), c, fb)
+
+  private[transformers] def ngrams(a: Seq[String]): Seq[String] = {
+    val ngrams = for (i <- low to high) yield a.sliding(i).map(_.mkString(separator))
```
andrewsmartin: Does it make sense to unroll this and try to optimize? Vocabulary space is obviously a problem for time complexity

core/src/test/scala/com/spotify/featran/transformers/NGramsTransformerSpec.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object NGramsTransformerSpec extends TransformerProp("NHotEncoder") {
+
+  private implicit val labelArb = Arbitrary {
+    Gen.choose(1, 5).flatMap(Gen.listOfN(_, Gen.alphaStr))
+  }
+
+  property("default") = Prop.forAll { xs: List[List[String]] =>
+    val transformer = new NGramsTransformer("n_gram", " ", 2, 4)
+    val ngrams = xs.map(transformer.ngrams(_))
```
andrewsmartin: I don't really like this test because it doesn't exercise the `ngrams` function independently, which seems harder to do in a property test. Ideally I'd have a separate unit test to check that in isolation. What do you think?

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+package com.spotify.featran.transformers
+
+import com.spotify.featran.FeatureBuilder
+
+import scala.collection.SortedMap
+
+/**
+ * Transform a collection of sentences, where each row is a `Seq[String]` of the words / tokens,
+ * into a collection containing all the n-grams that can be constructed from each row. The feature
+ * representation is an m-hot encoding (see [[NHotEncoder]]) constructed from an expanded vocabulary
```
nevillelyh: You mean "n-hot"?

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+ * of all of the generated n-grams.
+ *
+ * N-grams are generated based on a specified range of `low` to `high` (inclusive) and are joined by
+ * the given `separator` (the default is ""). For example, an [[NGramsTransformer]] with
```
nevillelyh: I'd use space " " as default, since you might run into duplicates with simple concat in typical NLP application.

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+ * `separator = ""`, `low = 2`, and `high = 3`, the row `["a", "b", "c", "d", "e"]` would produce
+ * the ngrams `["ab", "bc", "cd", "de", "abc", "bcd", "cde"]`.
+ *
+ * As with [[NHotEncoder]], missing values are transformed to [0.0, 0.0, ...].
+ */
+object NGramsTransformer {
```
nevillelyh: No other transformer actually have `Transformer` in the name, just call it `NGrams`?

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
```diff
@@ -0,0 +1,68 @@
...
+  type TfType = Transformer[Seq[String], Set[String], SortedMap[String, Int]]
+
+  /**
+   * Create a new [[NGramsTransformer]] instance.
+   * @param separator a string separator used to join individual tokens
+   * @param low the smallest size of the generated *-grams
+   * @param high the largest size of the generated *-grams
+   */
+  def apply(name: String, separator: String, low: Int, high: Int): TfType =
```
nevillelyh: `def apply(name: String, low: Int = 1, high: Int = -1, sep: String = " ")`? This way 1 `apply` handles all. Just need to make it clear that by default it generates n-grams all the way to `n = xs.length` and may be expensive.

core/src/main/scala/com/spotify/featran/transformers/NGramsTransformer.scala
nevillelyh: Not necessarily but maybe worth making it lazy with `a.toStream` etc.

core/src/test/scala/com/spotify/featran/transformers/NGramsTransformerSpec.scala
nevillelyh: It's OK for now. We can make assertions more black box as part of #53.

andrewsmartin:
@nevillelyh addressed all your comments, feel free to take another look and lmk if good to merge

PR #59, opened 2017-10-23
Fix concurrency problem in java and sparse arrays

ravwojdyla:
Memoizing sparse array feature builder fail in concurrent environment
because that builder holds state.


nevillelyh:
Rather than handcrafting test specific for one `FeatureBuilder`, make it more generic by moving it here?
https://github.com/spotify/featran/blob/master/core/src/test/scala/com/spotify/featran/FeatureBuilderSpec.scala#L30

We can make `fb: FeatureBuilder[F]` a context bound, and verify that two invocation of `implicitly[FeatureBuilder[F]]` return different instances.

nevillelyh:
Actually I was wrong. We can't cover it in `FeatureBuildSpec` as this is specific to Java due to the lack of HKTs.

PR #62, opened 2017-10-26
benchmarking for Hot Encoders

annaks:
Added benchmarking for One-Hot and N-Hot encoders when researching performance issues.  Ended up being related to transforming `Option[String]` to `List[String]` when using N-Hot which wasnt needed in One-Hot.  

Pruned tests for relevant output:
```
> jmh:run  -i 40 -wi 20 -f1 -t1
...
[info] Benchmark                                                            Mode  Cnt      Score      Error  Units
[info] HotEncoderTransformersBenchmark.nHotOptionalEncoder                  avgt   40    164.940 ±    8.714  ns/op
[info] HotEncoderTransformersBenchmark.nHotOptionalEncoder1000Values        avgt   40    148.615 ±    5.887  ns/op
[info] HotEncoderTransformersBenchmark.nHotRequiredEncoder                  avgt   40    157.949 ±   13.033  ns/op
[info] HotEncoderTransformersBenchmark.nHotRequiredEncoder1000Values        avgt   40    146.528 ±    3.136  ns/op
[info] HotEncoderTransformersBenchmark.nHotRequiredWithListMatch            avgt   40    277.947 ±   23.363  ns/op
[info] HotEncoderTransformersBenchmark.nHotRequiredWithListMatch1000Values  avgt   40  93387.703 ± 1969.640  ns/op
[info] HotEncoderTransformersBenchmark.oneHotOptionalEncoder                avgt   40    305.788 ±   45.427  ns/op
[info] HotEncoderTransformersBenchmark.oneHotOptionalEncoder1000Values      avgt   40    252.499 ±   34.495  ns/op
[info] HotEncoderTransformersBenchmark.oneHotRequiredEncoder                avgt   40    630.324 ±  101.491  ns/op
[info] HotEncoderTransformersBenchmark.oneHotRequiredEncoder1000Values      avgt   40    405.711 ±   99.847  ns/op
```


featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,193 @@
...
+    featureSpec.extract(input)
+  }
+
+  private val thisString = Gen.alphaStr.toString
+
+  // Lists of Lists for N-Hot
+  private implicit val arbitraryListString: List[List[String]] = List(
+    List(Gen.alphaStr.toString))
+  private implicit val arbitraryListString10Values: List[List[String]] = List(
+    Range(0, 10).map(_ => Gen.alphaStr.toString).toList)
```
nevillelyh: Not sure `Gen[T]#toString` does what you want. `def sample: Option[T]` is probably what you want (generates another random sample).
I'd just ditch scalacheck and hardcode all input to make it more predictable.

build.sbt
```diff
@@ -220,3 +220,19 @@ lazy val tensorflow: Project = Project(
...
+lazy val featranJmh: Project = Project(
+  "featran-jmh",
+  file("featran-jmh")
+).settings(
+  commonSettings ++ noPublishSettings,
+  description := "Featran JMH Microbenchmarks",
+  sourceDirectory in Jmh := (sourceDirectory in Test).value,
+  classDirectory in Jmh := (classDirectory in Test).value,
+  dependencyClasspath in Jmh := (dependencyClasspath in Test).value,
+  libraryDependencies ++= Seq(
```
nevillelyh: You can get rid of dependencies if inputs are hard coded.

build.sbt
```diff
@@ -220,3 +220,19 @@ lazy val tensorflow: Project = Project(
...
+    "org.scalacheck" %% "scalacheck" % scalacheckVersion % "test"
+  )
+).dependsOn(
+  core
+).enablePlugins(JmhPlugin)
```
nevillelyh: Minor: `\n` after last line.

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+  lazy private val data: Iterator[String] =
+    Range(0, 100000).map(i => s"string$i").toIterator
+  lazy private val thisString = data.toList.head
+
+  val fixedSeed = 235711L
+  def rngSeed: JRandom = new JRandom(fixedSeed)
+  rngSeed.setSeed(fixedSeed)
+
+  // Sampling without replacement (bernoulli)
+  def sample[T](data: Iterator[T], f: Double): Iterator[T] = {
```
annaks: @nevillelyh like this?

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  def oneHotRequired(input: List[String]): FeatureExtractor[List, String] = {
+    val featureSpec = FeatureSpec
+      .of[String]
+      .required(identity)(OneHotEncoder("one_hot"))
+    featureSpec.extract(input)
+  }
+
+  // Lists of Lists for N-Hot
+  private implicit val arbitraryListString: List[List[String]] = List(
```
nevillelyh: these don't need to be `implicit` any more right?

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+
+  def nHotOptional(
+      input: List[List[String]]): FeatureExtractor[List, List[String]] = {
+    val featureSpec = FeatureSpec
+      .of[List[String]]
+      .optional {
+        case List("") => None
+        case i        => Option(i)
+      }(NHotEncoder("n_hot"))
+    featureSpec.extract(input)
```
nevillelyh: I'd move `val featureSpec` out of these methods so they're not measured in the benchmark.
IIRC `.extract()` returns a `FeatureExtractor`, but doesn't actually perform any computation until you call `.featureValues`?

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+      .of[String]
+      .required(identity)(OneHotEncoder("one_hot"))
+    featureSpec.extract(input)
+  }
+
+  // Lists of Lists for N-Hot
+  private implicit val arbitraryListString: List[List[String]] = List(
+    List(thisString) ++ Seq(""))
+  private implicit val arbitraryListString1000Values: List[List[String]] = List(
+    sample(data, 0.01).toList)
```
nevillelyh: This may not return exactly 10000 values right?
I'd do something like this:

```
val vocab = (1 to 100).map("label" + _)
val s1000 = Seq.fill(1000)(vocab(rng.nextInt(100)) // pick 1
```

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+  private implicit val arbitraryListOptional100000ValuesWR
+    : List[Option[String]] =
+    sample(data, 0.99).map(i => Option(i)).toList ++ Range(1, 10).map(_ => None)
+
+  lazy private val data: Iterator[String] =
+    Range(0, 100000).map(i => s"string$i").toIterator
+  lazy private val thisString = data.toList.head
+
+  val fixedSeed = 235711L
+  def rngSeed: JRandom = new JRandom(fixedSeed)
```
nevillelyh: I'd call this `rng` not `rngSeed` since it's not really a seed?

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
```diff
@@ -0,0 +1,234 @@
...
+    List(Option(thisString))
+  private implicit val arbitraryListOptional1000Values: List[Option[String]] =
+    sample(data, 0.01).map(i => Option(i)).toList ++ Seq(None)
+  private implicit val arbitraryListOptional1000ValuesWR: List[Option[String]] =
+    sampleWR(data, 0.01).map(i => Option(i)).toList ++ Seq(None)
+  private implicit val arbitraryListOptional100000Values: List[Option[String]] =
+    sample(data, 0.99).map(i => Option(i)).toList ++ Seq(None)
+  private implicit val arbitraryListOptional100000ValuesWR
+    : List[Option[String]] =
+    sample(data, 0.99).map(i => Option(i)).toList ++ Range(1, 10).map(_ => None)
```
nevillelyh: `Range(1, 10).map(_ => None)` can also be `Seq.fill(10)(None)`

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
annaks: its reseeding the `rng` in both `def sample` and `def sampleWR`. but it might not be necessary to do this?

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
annaks: That makes a lot more sense - I'm seeing significant differences in benchmarks with more/less values when calling `.featureValues`.

featran-jmh/src/test/scala/com/spotify/featran/jmh/HotEncoderTransformersBenchmark.scala
annaks: got it - using the approach about with `vocab` and `s1000`, it makes sense that this is `rng` and not `rngSeed` :)

nevillelyh:
@annaks before you do any significant change:
"Actually can we rethink this a bit? I'd like a more generic wiring that covers all transformers, similar to"
https://github.com/spotify/featran/blob/master/core/src/test/scala/com/spotify/featran/Fixtures.scala

annaks:
@nevillelyh makes sense to make it more generic. Not sure if i catch the similarities with `Fixtures.scala`.  Do you have an idea for approach or where to start?

nevillelyh:
I got something working and pushed to this branch. Not sure if we need to cover all transformer use cases, e.g. match/required/optional, but rather a holistic view of relative transformer performances. Let's start there and fine tune later.

nevillelyh:
Preliminary benchmark result
```
[info] Benchmark                                 Mode  Cnt           Score          Error  Units
[info] TransformerBenchmark.binarizer            avgt   10      481269.503 ±   131886.180  ns/op
[info] TransformerBenchmark.bucketizer           avgt   10      668928.554 ±    28817.695  ns/op
[info] TransformerBenchmark.hashOneHotEncoder    avgt   10     2849103.162 ±   392180.220  ns/op
[info] TransformerBenchmark.heavyHitters         avgt   10   244559715.580 ± 21992300.154  ns/op
[info] TransformerBenchmark.identityB            avgt   10      477569.075 ±    27511.960  ns/op
[info] TransformerBenchmark.maxAbsScaler         avgt   10      494068.225 ±    68598.806  ns/op
[info] TransformerBenchmark.minMaxScaler         avgt   10      603591.981 ±    42558.904  ns/op
[info] TransformerBenchmark.nGrams               avgt   10  2548346192.500 ± 85579910.025  ns/op
[info] TransformerBenchmark.nHotEncoder          avgt   10    10538570.045 ±  1130112.637  ns/op
[info] TransformerBenchmark.nHotWeightedEncoder  avgt   10    14146653.134 ±   504867.825  ns/op
[info] TransformerBenchmark.normalizer           avgt   10     2705978.246 ±    52704.748  ns/op
[info] TransformerBenchmark.oneHotEncoder        avgt   10     1058843.721 ±    19560.361  ns/op
[info] TransformerBenchmark.polynomialExpansion  avgt   10    14376180.143 ±   236645.804  ns/op
[info] TransformerBenchmark.quantileDiscretizer  avgt   10     1652412.328 ±   217472.065  ns/op
[info] TransformerBenchmark.standardScaler       avgt   10      858217.772 ±    64037.554  ns/op
[info] TransformerBenchmark.vectorIdentity       avgt   10     3132018.992 ±   114528.851  ns/op
[info] TransformerBenchmark.vonMisesEvaluator    avgt   10     3501918.360 ±   421043.316  ns/op
```

annaks:
@nevillelyh nice!

PR #63, opened 2017-10-26
optimize for speed

slhansen:
When using featran in an apollo service we found that `NHotEncoder` was causing the feature extraction time to take up to 3 seconds. We profiled the code and found that these two functions:

https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala#L47-#L48

https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala#L63-#L64

took most of the time. The PR removes the call `filterKeys` and limits the number of times set subraction is used. I've attached a graph of how the latency times changed once `NHotEncoder` was removed.

featran_extraction_time

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
@@ -19,6 +19,7 @@ package com.spotify.featran.transformers
 
 import com.spotify.featran.{FeatureBuilder, FeatureRejection}
 
+import scala.collection.mutable.{Set => MSet}
 import scala.collection.SortedMap
 
 /**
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
```diff
@@ -19,20 +19,21 @@ package com.spotify.featran.transformers
...
 
 /**
- * Transform a collection of categorical features to binary columns, with at most N one-values.
- *
- * Missing values are transformed to [0.0, 0.0, ...].
- *
- * When using aggregated feature summary from a previous session, unseen labels are ignored and
- * [[FeatureRejection.Unseen]] rejections are reported.
- */
+  * Transform a collection of categorical features to binary columns, with at most N one-values.
```
nevillelyh: Can you keep the comment indentation as is please?

PR #66, opened 2017-11-07
Add quantile and IQR outlier rejectors

martinbomio:
Create transformers that will reject outliers based on:
  - First/Last quantile
  - IQR

The transformers don't modify the input feature, they just reject the ones that are outside the calculated bounds.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
+++ b/core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
@@ -32,6 +32,7 @@ object FeatureRejection {
   case class OutOfBound(lower: Double, upper: Double, actual: Double) extends FeatureRejection
   case class Unseen(labels: Set[String]) extends FeatureRejection
   case class WrongDimension(expected: Int, actual: Int) extends FeatureRejection
+  case class Outlier(actual: Double) extends FeatureRejection
 }
 
 /**
--- a/core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
@@ -1,0 +1,76 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.transformers
+
+import com.spotify.featran.FeatureRejection
+import com.twitter.algebird.QTreeAggregator
+
+ /**
+  * Rejects features if they fall outside of either 1.5 times the IQR below the first quartile,
+  * or [[IQROutlierRejector.DefaultFactor]] times the IQR above the third.
+  *
+  * The IQR or inter quartile range is the range between the first and the third quartiles.
+  *
+  * The bin ranges are chosen using the Algebird's QTree approximate data structure. The precision
+  * of the approximation can be controlled with the `k` parameter.
+  *
+  * All values are transformed to zero vectors.
+  *
+  * Values contained in the last bucket will report [[FeatureRejection.Outlier]] as rejection.
+  *
+  * When using aggregated feature summary from a previous session, values outside of previously seen
+  * `[min, max]` will also report [[FeatureRejection.Outlier]] as rejection.
+  */
+object IQROutlierRejector {
+   private val DefaultFactor = 1.5
+
+   /**
+    * Create a new [[IQROutlierRejector]] instance.
+    *
+    * @param rejectLower whether to reject outlier based on the lower igr bound.
+    * @param rejectUpper whether to reject outlier based on the upper iqr bound.
+    * @param k           precision of the underlying Algebird QTree approximation
+    */
+  def apply(name: String,
+            rejectLower: Boolean = true,
+            rejectUpper: Boolean = true,
+            k: Int = QTreeAggregator.DefaultK,
+            factor: Double = DefaultFactor)
+  : Transformer[Double, BaseQuantileRejector.B, BaseQuantileRejector.C] =
+    new IQROutlierRejector(name, rejectLower, rejectUpper, k, factor)
+
+}
+
+private class IQROutlierRejector(name: String,
+                                 rejectLower: Boolean,
+                                 rejectUpper: Boolean,
+                                 k: Int,
+                                 val factor: Double)
+  extends QuantileOutlierRejector(name, rejectLower, rejectUpper, 4, k) {
+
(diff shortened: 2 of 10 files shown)


core/src/main/scala/com/spotify/featran/transformers/QuantileOutlierRejector.scala
```diff
@@ -0,0 +1,136 @@
...
+  // scalastyle:off cyclomatic.complexity
+  override def buildFeatures(a: Option[Double], c: C, fb: FeatureBuilder[_]): Unit =
+    a match {
+      case Some(x) =>
+        val (fq, lq, min, max) = c
+        val (l, u) = calculateBounds(fq, lq)
+        fb.add(name, x)
+        // all elements can't be the same
+        if (x < min || x > max){
+          fb.reject(this, FeatureRejection.Outlier(x))
```
martinbomio: decided to reject as `Outlier` if they are outside the seen boundary but could easily change it to `OutOfBound`

core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
```diff
@@ -31,6 +31,7 @@ object FeatureRejection {
   case class OutOfBound(lower: Double, upper: Double, actual: Double) extends FeatureRejection
   case class Unseen(labels: Set[String]) extends FeatureRejection
   case class WrongDimension(expected: Int, actual: Int) extends FeatureRejection
+  case class Outlier(actual: Double) extends FeatureRejection
```
martinbomio: could also add lower/upper bound here, was not sure to add it since I am not sure it will generalize to all kind of Outlier rejectors

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
```diff
@@ -0,0 +1,71 @@
...
+
+private class IQROutlierRejector(name: String,
+                          override val rejectLower: Boolean,
+                          override val rejectUpper: Boolean,
+                          override val k: Int)
+  extends QuantileOutlierRejector(name, rejectLower, rejectUpper, 4, k) {
+
+  override def calculateBounds(fq: Double, lq: Double): (Double, Double) = {
+    val iqr = lq - fq
+    val l = fq - (iqr * 1.5)
```
martinbomio: haven't seen this factor being tuned, but could be a parameter of the transformer

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
```diff
@@ -0,0 +1,71 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.transformers
+
+import com.spotify.featran.FeatureRejection
+import com.twitter.algebird.QTreeAggregator
+
+/**
```
nevillelyh: Can you use javadoc style? See this:
https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
```diff
@@ -0,0 +1,71 @@
...
+  * Transform features by passing them through, and rejecting them if they
+  * fall outside of either 1.5 times the IQR below the first quartile,
+  * or 1.5 times the IQR above the third.
+  *
+  * The IGR or inter quartile range is the range between the first and the third quartiles.
```
nevillelyh: Should be IQR?

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
```diff
@@ -0,0 +1,71 @@
...
+            rejectLower: Boolean = true,
+            rejectUpper: Boolean = true,
+            k: Int = QTreeAggregator.DefaultK)
+  : Transformer[Double, BaseQuantileRejector.B, BaseQuantileRejector.C] =
+    new IQROutlierRejector(name, rejectLower, rejectUpper, k)
+
+}
+
+private class IQROutlierRejector(name: String,
+                          override val rejectLower: Boolean,
```
nevillelyh: Indentation. Also since you're passing these params to super constructor, you can remove the `override val`.

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
nevillelyh: Yeah it should at least be tuneable. What's the reasoning for 1.5 as default?

core/src/main/scala/com/spotify/featran/transformers/QuantileOutlierRejector.scala
```diff
@@ -0,0 +1,136 @@
...
+            numBuckets: Int = 3,
+            k: Int = QTreeAggregator.DefaultK)
+  : Transformer[Double, BaseQuantileRejector.B, BaseQuantileRejector.C] =
+    new QuantileOutlierRejector(name, rejectLower, rejectUpper, numBuckets, k)
+}
+
+private class QuantileOutlierRejector(name: String,
+                                      val rejectLower: Boolean,
+                                      val rejectUpper: Boolean,
+                                      override val numBuckets: Int,
```
nevillelyh: Remove `override val`

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
nevillelyh: This transformer is bypassed if both `rejectLower` & `rejectUpper` are false right? Should we add a check for that?

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
martinbomio: yeah, will make it a param. Seems like it is the most common value used for this technique.
https://math.stackexchange.com/questions/966331/why-john-tukey-set-1-5-iqr-to-detect-outliers-instead-of-1-or-2
```
We certainly CAN use whatever outlier bound we wish to use, but we will have to justify it eventually. In the not-so-recent past, it was typical to expect distributions to be Gaussian. With that assumption, ±1IQR is too exclusive, resulting in too MANY outliers, ±2IQR is too inclusive, resulting in too FEW outliers. ±1.5IQR is easy to remember, and is a reasonable compromise, under assumptions of Gaussianity.
```

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
martinbomio: yup, typo

core/src/main/scala/com/spotify/featran/transformers/IQROutlierRejector.scala
martinbomio: it is checked in `QuantileOutlierRejector`

nevillelyh:
@richwhitjr wanna take a look also?

richwhitjr:
The code itself looks good but I am worried about the current API with a rejection only transformer.  I understand why this is needed and it is useful but I am not sure if end users will understand that adding this transformer will reject the entire record.  Maybe we need to think of a new api for these types of filters?  @nevillelyh

nevillelyh:
@richwhitjr can you think of other "rejection only" use cases? If not I'm fine with merging as is and add "rejection only" semantics later when the need arises.

richwhitjr:
Sure lets merge for now, we can refactor in the future.

PR #69, opened 2017-11-18
add RecordExtractor, fix #68

### LINKED ISSUE #68: Optimize extract with settings for streaming/backend use cases
Right now `FeatureSpec#extractWithSettings` does 2 things, parsing JSON settings and extract with `.map`. This could be inefficient in a backend where data is `Seq[T]` of 1 element.

We should either split it into 2 steps or add a streaming API so that elements can be feed into the `Seq` lazily.

nevillelyh:


core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -99,3 +99,66 @@ class FeatureExtractor[M[_]: CollectionType, T] private[featran]
...
+  /**
+   * Extract feature values from a single record, with values in the same order as names in
+   * [[featureNames]] with rejections keyed on feature name and the original input record.
+   */
+  def featureResult(record: T): FeatureResult[F, T] = synchronized {
+    input.feed(record)
+    output.next()
+  }
+
+  private class PipeIterator extends Iterator[T] {
```
richwhitjr: Is this different than `iterator.buffered`?

nevillelyh: No, `BufferedIterator` just allows you to `head` without advancing the element. This is technically not an `Iterator`, I just can't think of a better word.

nevillelyh:
It makes less sense in `MultiFeatureSpec` I think? That's mostly intended for data pipelines so you can do single pass with multiple specs. The `RecordExtractor` is more useful in streaming/backends when you only run 1 single spec.

richwhitjr:
Only case I could think of right now is for mini-batch learning where you want to use MultiSpec to pull apart labels and features to update a model.  We can see if someone needs that though in the future.

nevillelyh:
Yeah we can implement those in `MultiFeatureSpec` later, either by duplicating code or abstract logic away in a trait/HKT, not sure how doable though.

BTW benchmark result processing `Seq` of 10 items and single items:

```
[info] # Run complete. Total time: 00:01:22
[info] Benchmark                       Mode  Cnt      Score     Error  Units
[info] ExtractorBenchmark.collection   avgt   10  23038.389 ± 169.491  ns/op
[info] ExtractorBenchmark.collection1  avgt   10  17123.628 ± 629.408  ns/op
[info] ExtractorBenchmark.record       avgt   10   4897.009 ± 133.275  ns/op
[info] ExtractorBenchmark.record1      avgt   10    635.447 ±   5.899  ns/op
```

PR #70, opened 2017-11-27
Adding functionality for java feature crosses

znichols:


richwhitjr:
LGTM, @nevillelyh should probably take a look also.  

Related note, we may want to benchmark these java calls in the future especially with crossing.  Wonder if there are some cheap performance improvements hidden in there.

PR #72, opened 2017-12-19
Feature Spec Extension

richwhitjr:
This method should hopefully make it easier to share feature specs when you have a spec that depends on some sub type of a larger object.  The idea is that you have some custom class that is specific to your domain you may have sub object that you share across projects.

The function in the method allows for the mapping between the higher level object and the reusable spec object.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FeatureSpec.scala
+++ b/core/src/main/scala/com/spotify/featran/FeatureSpec.scala
@@ -87,6 +87,21 @@ class FeatureSpec[T] private[featran] (private[featran] val features: Array[Feat
   }
 
   /**
+   * Extend this Spec with another Spec but allowing for a function to be called first.
+   * Useful for an existing Spec on a subobject.
+   * @param f Function to apply before calling the extending spec.
+   * @param spec Spec to apply.
+   * @tparam S Type of the extension spec
+   */
+  def extend[S](f: T => S)(spec: FeatureSpec[S]): FeatureSpec[T] = {
+    val composedFeatures = spec.features.map{feature =>
+      val t = feature.transformer.asInstanceOf[Transformer[Any, _, _]]
+      new Feature(f.andThen(feature.f), feature.default, t)
+    }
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/spotify/featran/FeatureSpec.scala
```diff
@@ -86,6 +86,21 @@ class FeatureSpec[T] private[featran] (private[featran] val features: Array[Feat
...
   }
 
+  /**
+   * Extend this Spec with another Spec but allowing for a function to be called first.
+   * Useful for an existing Spec on a subobject.
+   * @param f Function to apply before calling the extending spec.
+   * @param spec Spec to apply.
+   * @tparam S Type of the extension spec
+   */
+  def extend[S](f: T => S)(spec: FeatureSpec[S]): FeatureSpec[T] = {
```
nevillelyh: This looks like a `specA.map[B](f: A => B)` followed by a `MultiFeatureSpec(specA, specB)`? Is that a better modular approach?

richwhitjr: Not exactly a multispec.  I don't really want a separate list of vectors on put here instead I want to reuse some specs but still get out a single vector as normal.  Could be more difficult to deal with the multiple vectors and names in that case.

So if you have a feature objects like 

```
case class House(sqft: Double, color: String)
case class HousePrice(house: House, price: Float)
case class HouseSold(house: House, sold: Boolean)
```

It would be nice to write one House spec and use for both price and sold without needing to rewrite code and worry about multiple splits.

nevillelyh: Oh I get now. What about this?
`def compose(spec: FeatureSpec[S])(f: T => S)`

`compose` feels more natural (using another spec as part of this), and `f` as 2nd arg so it's easy to write multi-line fn.

richwhitjr: Makes sense, let me change it to that.  I was debating between compose and extend.

PR #73, opened 2017-12-26
In Memory MDL Transformer

richwhitjr:
This PR implements a MDLP transformer for continuous values.  Given a label and a value the transformer finds the entropy split to find the optimal buckets.  This can take a lot of the guess work out of figuring out the number of buckets compared to quantiles.

Because MDL requires iteration the values are pulled into memory.  There is an optional sample rate argument that can be set to lower the amount of data pulled into memory.  Also because sumOption is not implemented yet in Featran the values are accumulated using the semigroup plus which can be inefficient.


core/src/main/scala/com/spotify/featran/transformers/mdl/ThresholdFinder.scala
```diff
@@ -0,0 +1,158 @@
...
+    val diff = bucketInfo.k * bucketInfo.hs - k1 * hs1 - k2 * hs2
+    val delta = log2(math.pow(3, bucketInfo.k) - 2) - diff
+    (gain - (log2(bucketInfo.s - 1) + delta) / bucketInfo.s, weightedHs, s1, s2)
+  }
+
+  /**
+    * Evaluates boundary points and selects the most relevant candidates (sequential version).
+    * Here, the evaluation is bounded by partition as the number of points is small enough.
+    *
+    * @param candidates RDD of candidates points (point, class histogram).
```
yonromai: shouldn't keep the `RDD`?

core/src/main/scala/com/spotify/featran/transformers/MDL.scala
```diff
@@ -0,0 +1,143 @@
...
+ * This Transformer computes the optimal number of buckets using
+ * minimum description length. That is an entropy measurement between
+ * the values and the tagets.
+ *
+ * The Transformer expects an MDLRecord where the first field is a label
+ * and the second value is the scalar that will be transformed into buckets.
+ *
+ * MDL is an iterative algorithm so all of the data needed to compute the buckets
+ * will need to be pulled into memory.  If you run into memory issues the sampleRate
+ * parameter will need to be lowered.
```
yonromai: do you have external pointers about this version MDL? would love to read more about it

richwhitjr: The core part of the code comes from this repo.  https://github.com/sramirez/spark-MDLP-discretization.  

Old school paper: http://sci2s.ugr.es/keel/pdf/algorithm/congreso/fayyad1993.pdf

yonromai: nice! will take a look - if anyone else is interested, there is a cleaner scan of the same paper here: http://web.donga.ac.kr/kjunwoo/files/Multi%20interval%20discretization%20of%20continuous%20valued%20attributes%20for%20classification%20learning.pdf

core/src/main/scala/com/spotify/featran/transformers/mdl/ThresholdFinder.scala
yonromai: @richwhitjr ^

yonromai:
besides the minor comments, LGTM. @nevillelyh @ravwojdyla wanna double check?

core/src/test/scala/com/spotify/featran/transformers/mdl/ThresholdFinderTest.scala
```diff
@@ -0,0 +1,43 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.transformers.mdl
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class ThresholdFinderTest extends FlatSpec with Matchers {
+  it should "Run few values finder on nLabels = 3 feature len = 4" in {
```
nevillelyh: "ThresholdFinder" should "run ..." ?

core/src/test/scala/com/spotify/featran/transformers/MDLSpec.scala
```diff
@@ -0,0 +1,55 @@
...
+    for {
+      value <- Gen.oneOf(Seq("1", "2", "3"))
+      n <- Gen.posNum[Double]
+    } yield MDLRecord(value, n)
+  }
+
+  property("default") = Prop.forAll { xs: List[MDLRecord[String]] =>
+    val ranges = new MDLPDiscretizer[String](xs.map(l => (l.label, l.value))).discretize()
+
+    val slices = ranges.tail
```
nevillelyh: `ranges` is from `discretize()` and a `List[Double]`, so `slices` is a `List` of 1 single `Double, and `names` is always `Seq("mdl_0")`. Is that correct?

nevillelyh: Nevermind my brain froze. `tail` is `n - 1`.

PR #75, opened 2017-12-30
Add support for TFRecordSpec

yonromai:
**DO NOT MERGE YET: https://github.com/spotify/scio/pull/990 should be merged first.**

Add `TFRecordSpec` support for both `FeatureSpec` and `MultiFeatureSpec`.

Enables this convient API:
```scala
val dataset = MultiFeatureSpec(features, label)
  .extract(collection)

val (train, test) = dataset
  .featureValues[tf.Example]
  .randomSplit(.9)

train.saveAsTfExampleFile(args("output") + "/train", dataset)
test.saveAsTfExampleFile(args("output") + "/test", dataset)
```

**Sibling PR: https://github.com/spotify/spotify-tensorflow/pull/13**

### DIFF
--- a/scio/src/main/scala/com/spotify/featran/scio/TFExampleScollectionFunctions.scala
+++ b/scio/src/main/scala/com/spotify/featran/scio/TFExampleScollectionFunctions.scala
@@ -1,0 +1,120 @@
+/*
+ * Copyright 2017 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.featran.scio
+
+import com.spotify.featran.{FeatureExtractor, MultiFeatureExtractor}
+import com.spotify.scio.io.Tap
+import com.spotify.scio.tensorflow._
+import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.io.Compression
+import org.tensorflow.{example => tf}
+import scala.concurrent.Future
+
+private object FeatranTFRecordSpec {
+
+  def fromFeatureSpec(featureNames: SCollection[Seq[String]]): TFRecordSpec = {
+    TFRecordSpec.fromSColSeqFeatureInfo(
+      featureNames.map(_.map(n => FeatureInfo(n, FeatureKind.FloatList, Map()))))
+  }
+
+  def fromMultiSpec(featureNames: SCollection[Seq[Seq[String]]]): TFRecordSpec = {
+    val featureInfos = featureNames.map(_.zipWithIndex.flatMap {
+      case (sss, i) => sss.map(n =>
+        FeatureInfo(n, FeatureKind.FloatList, Map("multispec-id" -> i.toString)))
+    })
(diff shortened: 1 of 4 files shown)


yonromai:
Once https://github.com/spotify/scio/pull/990 has been merged and a new version of Scio has been released, I'll update the `build.sbt` to not depend on `SNAPSHOT`.

yonromai:
@ravwojdyla @nevillelyh PTAL

scio/src/main/scala/com/spotify/featran/scio/package.scala
```diff
@@ -18,19 +18,37 @@
...
    * [[CollectionType]] for extraction from Scio `SCollection` type.
    */
   implicit object ScioCollectionType extends CollectionType[SCollection] {
     override def map[A, B: ClassTag](ma: SCollection[A], f: (A) => B): SCollection[B] = ma.map(f)
+
     override def reduce[A](ma: SCollection[A], f: (A, A) => A): SCollection[A] = ma.reduce(f)
+
     override def cross[A, B: ClassTag](ma: SCollection[A],
                                        mb: SCollection[B]): SCollection[(A, B)] = ma.cross(mb)
+
```
ravwojdyla: `-`

scio/src/main/scala/com/spotify/featran/scio/package.scala
```diff
@@ -18,19 +18,37 @@
 package com.spotify.featran
 
 import com.spotify.scio.values.SCollection
+import org.tensorflow.{example => tf}
 
 import scala.reflect.ClassTag
 
+
```
ravwojdyla: `-`

scio/src/main/scala/com/spotify/featran/scio/TFExampleScollectionFunctions.scala
```diff
@@ -0,0 +1,123 @@
...
+import com.spotify.featran.{FeatureExtractor, MultiFeatureExtractor}
+import com.spotify.scio.ScioContext
+import com.spotify.scio.io.Tap
+import com.spotify.scio.tensorflow._
+import com.spotify.scio.values.SCollection
+import org.apache.beam.sdk.io.Compression
+import org.tensorflow.{example => tf}
+
+import scala.concurrent.Future
+
```
ravwojdyla: `-`

scio/src/main/scala/com/spotify/featran/scio/TFExampleScollectionFunctions.scala
```diff
@@ -0,0 +1,123 @@
...
+  def fromFeatureSpec(featureNames: SCollection[Seq[String]]): TFRecordSpec = {
+    TFRecordSpec.fromSColSeqFeatureInfo(
+      featureNames.map(_.map(n => FeatureInfo(n, FeatureKind.FloatList, Map()))))
+  }
+
+  def fromMultiSpec(featureNames: SCollection[Seq[Seq[String]]]): TFRecordSpec = {
+    val featureInfos = featureNames.map(_.zipWithIndex.flatMap {
+      case (sss, i) => sss.map(n =>
+        FeatureInfo(n, FeatureKind.FloatList, Map("multispec-id" -> i.toString)))
```
ravwojdyla: Just want to make sure if I understand this, so this `Map` might as well be a `Option[String]` right?

yonromai: Right, the way this is right now, there is either 0 entries or 1 entry in the Map, so it did not have to be a Map.
It's a bit of speculative engineering, but I thought it would be convenient to have a way to tag feature with whatever we want. This provides much more flexibility to TFRecordSpec.

scio/src/main/scala/com/spotify/featran/scio/package.scala
yonromai: @ravwojdyla  all of this is Intellij code format feature. Do you have different settings or do you not use it?

nevillelyh:
Talked offline and this should be part of `scio-tensorflow`.

PR #81, opened 2018-02-16
add Java TF Example extractor

nevillelyh:


java/src/main/java/com/spotify/featran/java/JFeatureExtractor.java
```diff
@@ -79,4 +80,11 @@ public String featureSettings() {
...
 
-}
+  /**
+   * Java wrapper for {@link FeatureExtractor#featureValues(FeatureBuilder, ClassTag)}.
+   */
+  public List<Example> featureValuesExample() {
+    return JavaOps.featureValuesExample(self);
+  }
+
+}
```
ravwojdyla: `+` newline?

nevillelyh: DAMN IntelliJ

nevillelyh:
This will pull in TF proto dependency though (right now 1.4.0) but it should be stable enough and not cause conflicts?

nevillelyh:
OTOH none the TF artifacts, e.g. `org.tensorflow:{proto,tensorflow,libtensorflow}` has any transitive dependency. Maybe a plus side of them using bazel?

ravwojdyla:
@nevillelyh is it only for the `tf.Example`? we could also make it `provided`?

nevillelyh:
I think that will break those not using `tf.Example` output and without an explicit dependency, since the classloader will complain when loading.
Keep it in and let users override version (easy)?

ravwojdyla:
ah, yes - i guess it will, sure.

PR #82, opened 2018-02-28
Allow access to internal FeatureSpec

ravwojdyla:


nevillelyh:
😂 was just thinking about this.

Issue #83, opened 2018-02-28
Support feature extraction in XGBoost formats


ravwojdyla:
We should add support for feature extraction in `XGBoost` formats: `LabeledPoint`, `DMatrix`, ?

richwhitjr:
LabeledPoint may make sense although I don't think it is serializable so that may be a problem?  We can attempt to solve for it though or just let users deal with it how they want.

DMatrix feels like more of an unknown.  Right now aside from numpy we write one row into one feature type.  Do you think we should save all of the rows into a single DMatrix or store each vector into a single row DMatrix?

ravwojdyla:
@richwhitjr I have not investigated this much, so I have few answers. I believe `LabeledPoint` is [serializable](https://github.com/dmlc/xgboost/blob/master/jvm-packages/xgboost4j/src/main/scala/ml/dmlc/xgboost4j/LabeledPoint.scala) (?)

richwhitjr:
Ah yep, missed that.

PR #84, opened 2018-02-28
Fixes: #83. Support XGBoost's LabeledPoint extraction.

ravwojdyla:


regadas:
@ravwojdyla Good stuff 👍

PR #86, opened 2018-03-09
Fix #85. Expose LabeledPoint in java API

### LINKED ISSUE #85: Expose xgboost feature extractor in java api

ravwojdyla:


nevillelyh:
2.12 build failed. Class loader fubar?
```
[info] com.spotify.featran.java.JavaTest *** ABORTED ***
[info]   java.lang.NoClassDefFoundError: scala/Product$class
[info]   at ml.dmlc.xgboost4j.LabeledPoint.<init>(LabeledPoint.scala:35)
```

nevillelyh:
I think it's due to the fact that `xgboost` is 2.11 only. Need to fix our `build.sbt` accordingly.

Issue #76, opened 2018-01-14
Upgrade breeze


nevillelyh:
1.0-RC1 is out. Hopefully a stable release soon.
https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/scala-breeze/8b5IGfo10rU/Z1HyEpTbAQAJ

ravwojdyla:
Fixed in #88

PR #88, opened 2018-03-12
Bump dependencies, still keep circe at 0.8.0

ravwojdyla:


ravwojdyla:
Fixes #76

ravwojdyla:
@nevillelyh PTAL

PR #92, opened 2018-03-14
PositionEncoder Transformer

### LINKED ISSUE #91: Add PositionalOneHotEncoder Transformer
Name needs to some work but the idea is when you have some categorical features like (red, blue, green) and have a feature that is green instead of output <0, 0, 1> the transformer should output 2.

richwhitjr:
Fixes #91.  General idea is that instead of outputting a binary vector the transformer instead outputs the position of the category in the sorted sets of possible values. The one small quirk of this though is that missing data will collide with the 0 position.  For labels the first class having a position of 0 seems standard so I don't think we want 1 index.

To work around collisions there is now a new rejection type called Collide that users can remove if they know they don't expect missing values in the category.

The thought is that this transformer would mostly be used around labels for supervised problems so removing rows with missing labels may be reasonable.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/PositionEncoder.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/PositionEncoder.scala
@@ -1,3 +1,20 @@
+/*
+ * Copyright 2018 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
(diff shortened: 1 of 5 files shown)


richwhitjr:
@nevillelyh @ravwojdyla

core/src/main/scala/com/spotify/featran/transformers/PositionEncoder.scala
```diff
@@ -0,0 +1,61 @@
...
+ * under the License.
+ */
+
+package com.spotify.featran.transformers
+
+import com.spotify.featran.{FeatureBuilder, FeatureRejection}
+
+import scala.collection.SortedMap
+
+/**
```
nevillelyh: Use javadoc style please.
https://github.com/spotify/scio/wiki/Style-Guide#documentation-style

```java
/**
 * xyz ...
 */
```

core/src/main/scala/com/spotify/featran/transformers/PositionEncoder.scala
```diff
@@ -0,0 +1,61 @@
...
+  * Transform a collection of categorical features to a single value that is the position
+  * of that feature within the complete set of categories.
+  *
+  * Missing values are transformed to zero. Thought may need to be done on how to reject these
```
nevillelyh: "Thought may need to be done"?

PR #93, opened 2018-03-14
add TopNOneHotEncoder, fix #89

### LINKED ISSUE #89: Support Selective OneHotEncoder
Currently the OHE encodes all the values, but there might be 1000s of them which will blow up the training data. It'll be good to have a control to limit the number of columns generated after OHE.

Some solutions from the discussions on slack:

1. Have a parameter "N" which defines the number of columns and most occurring N categories will be kept.
2. Have the list of required categories be defined in the feturespec. The user will define a list of N categories required and only those will be OHE, others ignored.
3. Have a percentile defined, e.g. keep all categories accounting for say 90% of the observations.

From initial discussions it seemed, #1 and #3 are harder to implement that #2? Open to discussion.

fallonchen:


core/src/main/scala/com/spotify/featran/transformers/TopNOneHotEncoder.scala
```diff
@@ -0,0 +1,118 @@
...
+  def apply(name: String, n: Int,
+            eps: Double = 0.001,
+            delta: Double = 0.001,
+            seed: Int = Random.nextInt)
+  : Transformer[String, SketchMap[String, Long], SortedMap[String, Int]] =
+    new TopNOneHotEncoder(name, n, eps, delta, seed)
+}
+
+private class TopNOneHotEncoder(name: String,
+                        val n: Int,
```
nevillelyh: Nit: indentation

core/src/main/scala/com/spotify/featran/transformers/TopNOneHotEncoder.scala
```diff
@@ -0,0 +1,118 @@
...
+          b += k -> r
+        }
+        b.result()
+      }
+
+  override def featureDimension(c: SortedMap[String, Int]): Int = c.size
+  override def featureNames(c: SortedMap[String, Int]): Seq[String] =
+    c.map(name + '_' + _._1)(scala.collection.breakOut)
+
+
```
nevillelyh: Nit: extra blank line

PR #97, opened 2018-03-30
upgrade circe and scio

nevillelyh:


scio/src/main/scala/com/spotify/featran/scio/FeatranKryoRegistrar.scala
```diff
@@ -0,0 +1,52 @@
...
+
+@KryoRegistrar
+class FeatranKryoRegistrar extends IKryoRegistrar {
+  override def apply(k: Kryo): Unit = {
+    k.forClass(new QTreeSerializer)
+  }
+}
+
+private class QTreeSerializer extends Serializer[QTree[Any]] {
+  setImmutable(true)
```
andrewsmartin: What is this for? From the doc it says `This causes {@link #copy(Kryo, Object)} to return the original object.`, is that called anywhere?

nevillelyh: Not in our code. But all `chill-algebird` serializers have this because most Algebird types are immutable by design.

PR #100, opened 2018-04-10
increase dim by 1 for encodeMissingValue

slhansen:
When `encodeMissingValue` is true the dimension of the features will be the number of entries in the sorted map plus 1 for `__missing__`


nevillelyh:
Just curious, the other encoders are not affected?
Also wondering if this can be caught in a test?

slhansen:
The unit tests use `.featureResults[Seq[Double]]` or `featureValues[Seq[Double]]`. I ran into an error when doing `featureValues[SparseVector[Double]]`. This `assert`

https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/FeatureBuilder.scala#L239

was failing.

nevillelyh:
Great catch. I'll add tests with `SparseVector` then to catch them in the future.

PR #102, opened 2018-04-16
Fix dependency conflicts

regadas:


build.sbt
```diff
@@ -123,8 +123,9 @@ lazy val core: Project = Project(
   moduleName := "featran-core",
   description := "Feature Transformers",
   libraryDependencies ++= Seq(
-    "com.twitter" %% "algebird-core" % algebirdVersion,
+    "com.twitter" %% "algebird-core" % algebirdVersion exclude ("org.typelevel", "algebra_2.11"),
```
nevillelyh: Which version are they depending on?

regadas: they are depending on `0.7.0` do you think is better to submit a PR on their side?

nevillelyh: We can do both, and revert our workaround once they fix it.

regadas: the main conflicts here is with `cats` and `shapeless`

nevillelyh: I think you can remove the `exclude ()` part, plus it only affects 2.11 while we also cross build for 2.12.

The explicit dependency down there should do it.

regadas: the exclude is actually required here ... good point on the cross build. Updated it to allow cross builds.

nevillelyh:
Format error now :)

PR #105, opened 2018-04-18
Add RecordExtractor state as a ThreadLocal

regadas:


core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -142,10 +159,7 @@ class RecordExtractor[T, F: FeatureBuilder: ClassTag] private[featran] (fs: Feat
    * Extract feature values from a single record, with values in the same order as names in
    * [[featureNames]] with rejections keyed on feature name and the original input record.
    */
-  def featureResult(record: T): FeatureResult[F, T] = synchronized {
-    input.feed(record)
-    output.next()
-  }
+  def featureResult(record: T): FeatureResult[F, T] = state.get().featureResult(record)
```
nevillelyh: Can we do some basic benchmarks?

core/src/test/scala/com/spotify/featran/CoverageSpec.scala
```diff
@@ -30,7 +30,8 @@ object CoverageSpec extends Properties("Coverage") {
     val fs = FeatureSpec.of[Double].required(identity)(Identity("id"))
     val settings = fs.extract(Seq(1.0)).featureSettings.head
     val e = fs.extractWithSettings[Seq[Double]](settings)
-    val f = e.getClass.getDeclaredField("iteratorCollectionType")
+    val f =
+      e.getClass.getDeclaredField("com$spotify$featran$RecordExtractor$$iteratorCollectionType")
```
nevillelyh: LOL

core/src/main/scala/com/spotify/featran/FeatureExtractor.scala
```diff
@@ -116,10 +116,27 @@ class RecordExtractor[T, F: FeatureBuilder: ClassTag] private[featran] (fs: Feat
...
+      input.feed(record)
+      output.next()
+    }
+  }
+
+  private val state = new ThreadLocal[State] {
+    override def initialValue(): State = {
+      val input: PipeIterator = new PipeIterator
+      val extractor: FeatureExtractor[Iterator, T] =
+        new FeatureExtractor[Iterator, T](fs, input, Some(Iterator.continually(settings)))
```
nevillelyh: This is the expensive part we're trying to avoid. This could still be a problem if we have a lot of short-lived threads (like in Dataflow streaming?). I wonder if there's a better way to solve this? Like making `FeatureExtractor` easily cloneable and share the parsed settings?

tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -17,26 +17,32 @@
...
+        .setFloatList(tf.FloatList.newBuilder().addValue(value.toFloat))
+        .build()
+      underlying.putFeature(name, feature)
+    }
+    override def skip(): Unit = Unit
+    override def skip(n: Int): Unit = Unit
+    override def result: tf.Example =
+      tf.Example.newBuilder().setFeatures(underlying).build()
+
+    override def newBuilder: FeatureBuilder[Example] = TensorFlowFeatureBuilder(underlying.clear())
```
nevillelyh: Are we reusing the `Features.Builder` in this case? That'll cause problem I think?

regadas: @nevillelyh yeah good point

PR #107, opened 2018-04-19
Closes #71: Make JMapWrapper final case class

### LINKED ISSUE #71: JMapWrapper should be serializable
Otherwise it won't work with Spark.

regadas:


regadas:
Dup see #108

PR #108, opened 2018-04-19
Closes #71: Make JMapWrapper final class

### LINKED ISSUE #71: JMapWrapper should be serializable

regadas:


core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
```diff
@@ -274,11 +274,14 @@ object FeatureBuilder {
 
 }
 
-private class JMapWrapper[K, V](val m: java.util.Map[K, V]) extends Map[K, V] {
+private final class JMapWrapper[K, V](m: java.util.Map[K, V]) extends Map[K, V] with Serializable {
```
nevillelyh: It's not a case class though, confusing commit message?

nevillelyh:
Do we need a test for this? IIRC it affects `featran-spark` when we extract features as `Map[String, T]`.

core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
regadas: @nevillelyh ah yeah updated

regadas:
@nevillelyh hm, yeah let's do that

regadas:
@nevillelyh test added

core/src/test/scala/com/spotify/featran/Fixtures.scala
```diff
@@ -43,6 +43,14 @@ object Fixtures {
                            Seq(0.0, 0.0, 0.0, 1.0, 0.0, 0.75),
                            Seq(0.0, 0.0, 0.0, 0.0, 1.0, 1.00))
 
+  val expectedMapValues = Seq(
+    Map("min_max" -> 0.0, "one_hot_a" -> 1.0),
+    Map("min_max" -> 0.25, "one_hot_b" -> 1.0),
+    Map("one_hot_c" -> 1.0, "min_max" -> 0.5),
```
nevillelyh: purely aesthetic, sort the keys so it's more readable?

PR #116, opened 2018-05-09
Revert #102

regadas:
After a little bit more digging around dependency resolution/conflicts ... I think it's wise to revert this change. The eviction `warn` that we get is only about a transitive dependency. If it was a direct one the story would be different.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -123,9 +123,8 @@ lazy val core: Project = Project(
   moduleName := "featran-core",
   description := "Feature Transformers",
   libraryDependencies ++= Seq(
-    "com.twitter" %% "algebird-core" % algebirdVersion excludeAll ("org.typelevel" %% "algebra"),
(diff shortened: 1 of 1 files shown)


build.sbt
```diff
@@ -123,9 +123,7 @@ lazy val core: Project = Project(
   moduleName := "featran-core",
   description := "Feature Transformers",
   libraryDependencies ++= Seq(
-    "com.twitter" %% "algebird-core" % algebirdVersion excludeAll ("org.typelevel" %% "algebra"),
-    "org.scalanlp" %% "breeze" % breezeVersion,
```
nevillelyh: Are we getting rid of `breeze` intentionally?

regadas: ah no ... just noticed that. i'll fix it

regadas: done

PR #123, opened 2018-05-23
Use @typeclass macro for CollectionType

regadas:


regadas:
after this one ... maybe we can apply the same strategy to `FeatureBuilder`

PR #130, opened 2018-05-31
treat empty input as a missing value

slhansen:
Want to treat the case where `xs` is an empty sequence as a missing value. In this case `keys` would also be empty.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
@@ -46,6 +46,16 @@ private class NHotEncoder(name: String, encodeMissingValue: Boolean)
 
   import MissingValue.missingValueToken
 
+  def checkForMissingValue(fb: FeatureBuilder[_],
+                           unseen: MSet[String],
+                           keys: Seq[String]): Unit = {
+    if (unseen.isEmpty && !keys.isEmpty) {
+      fb.skip()
+    } else {
+      fb.add(name + '_' + missingValueToken, 1.0)
+    }
+  }
+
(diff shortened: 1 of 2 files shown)


nevillelyh:
Travis failed format check, can you run `scalafmtCheck test:scalafmtCheck` locally to verify?

core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala
```diff
@@ -46,6 +46,16 @@ private class NHotEncoder(name: String, encodeMissingValue: Boolean)
 
   import MissingValue.missingValueToken
 
+  def checkForMissingValue(fb: FeatureBuilder[_],
```
regadas: @slhansen can we change the name of this method since it's actually modifying `FeatureBuilder`? `addMissingValue` perhaps?

slhansen: @nevillelyh it fails locally but won't give me a reason why :(
```
[info] Set current project to featran (in build file:/Users/Sammy/dev_clean/featran/)
[error] /Users/Sammy/dev_clean/featran/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala isn't formatted properly!
[error] (core / Compile / scalafmtCheck) /Users/Sammy/dev_clean/featran/core/src/main/scala/com/spotify/featran/transformers/NHotEncoder.scala isn't formatted properly!```

regadas: @slhansen you can do `sbt scalafmt` and it will format the code for you.

PR #137, opened 2018-06-06
expose transformer names to users

alaiacano:
The `features` are private within a `FeatureSpec` but it's useful to know all of the transformer names. 

For context, it came up because we have a situation like this where we want everything from one `FeatureSpec` in a `MultiSpec`, but just a subset of the other.

```scala
val labelSpec: FeatureSpec     // What we're trying to predict. We always want everything.
val featuresSpec: FeatureSpec  // features we're using. Might only want a subset.
val spec = MultiSpec(labelSpec, featuresSpec)

val transformersToUse = Set("transformer1", "transformer2") ++ labelSpec.transformerNames

spec
  .filter(f => transformersToUse.includes(f.transformer.name)
  .extract(data)
```


regadas:
@alaiacano on second thought, I was wondering if there's any reason to not do a filter on `featuresSpec` before creating a `MultiFeatureSpec`

```
val labelSpec: FeatureSpec     // What we're trying to predict. We always want everything.
val featuresSpec: FeatureSpec  // features we're using. Might only want a subset.
val transformersToUse = Set("transformer1", "transformer2") ++ labelSpec.transformerNames

val filteredFeaturesSpec = featuresSpec.filter(f => transformersToUse.contains(f.transformer.name)
val spec = MultiSpec(labelSpec, filteredFeaturesSpec)

spec.extract(data)
```

alaiacano:
@regadas Is there much of a difference to do the filtering before/after creating the `MultiSpec`? If it's just done once in the setup and not on each record that we try to extract, it should be approximately the same - is that correct?

Either way, we need to access the names of the transformers to do it. I wasn't sure if the list of features is private for a reason, so just exposed the names. Either solution works for me :)

PR #140, opened 2018-06-18
Slhansen/return names

slhansen:
The purpose of this PR is to return the names (e.g. `one_hot_<token>`) of the non-zero elements when calling `featureResults`. We need this information in order to experiment with field aware factorization machines.


nevillelyh:
Just curious what output type are you using for this particular use case? For types that store names, e.g. TF `Example` or `Map[String, Double]`, it's easy to get the list of names.

slhansen:
@nevillelyh the desired return type is a `SparseVector`.

nevillelyh:
Got it. My concern is this will add a lot of overhead to all `FeatureBuilder`s when the non-zero names aren't needed, since appending name per field per record is expensive. We can probably add a `NamedSparseVector` and corresponding `FeatureBuilder` instead for this specific case. WDYT?

nevillelyh:
An easier alternative is to call `extract` twice, once for `SparseVector`, once for `Map[String, Float]`, and use the `map.keys` to retrieve non-zero values.

nevillelyh:
Reworked with the named sparse array idea in #141.

regadas:
I think #141 is good to go! I think we can close this one?

PR #144, opened 2018-06-22
Fix #143: use filtered array to keep order

### LINKED ISSUE #143: Feature transformations order lost after filtering on a MultiFeatureSpec
We have noticed something strange when using the `filter` method of the `MultiFeatureSpec`: we are losing the order of the features as they were defined in the original `FeatureSpec` objects. We had a part of our pipeline that was actually refering explicitely to the feature number 0 and the feature number 2 of a resulting SparseVector:

```
val featureSpecs: FeatureSpec[AllFeatures] = FeatureSpec.of[AllFeatures]
    // ... Many feature transformations

val labelSpecs: FeatureSpec[AllFeatures] = FeatureSpec.of[AllFeatures]
    .optional(i => i.map(_.weight))(Identity("weight"))
    .optional(i => i.map(_.label))(Identity("label"))

val featureExtractor = MultiFeatureSpec(labelSpecs, featureSpecs)
           .filter{ f =>
            includedFeaturesWithLabels.contains(f.transformer.name)
          }
          .extract(rawFeatures)

def toTFTrainingExample(data: Seq[SparseVector[Float]]): Example = {
    // data is the list of the featurized raw features in the order of the FeatureSpec given
    // to multiSpecs
    val labels = data(0).data
    val features: SparseVector[Float] = data(1)
    val label = labels.headOption
    val weight = labels.length match {
      case 2 => Some(labels(1))
      case 1 => Some(1f)
      case _ => None
    }
    TensorFlowType[TrainingExample].toExample(
      TrainingExample(
        features.index.toList,
        features.data.toList,
        label,
        weight
      ))
}
```
But when the filtering is done, it seems the `Feature` are moved to a HashMap (losing the order at this point: https://github.com/spotify/featran/blob/master/core/src/main/scala/com/spotify/featran/MultiFeatureSpec.scala#L72-L76 then converted back to arrays.

That means we cannot assume the order of the features in the resulting `SparseVector`.

This looks like an unwanted behavior and definitely unexpected from the user point of view.

regadas:
Closes #143

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FeatureSpec.scala
+++ b/core/src/main/scala/com/spotify/featran/FeatureSpec.scala
@@ -123,13 +123,13 @@ class FeatureSpec[T] private[featran] (private[featran] val features: Array[Feat
    * @param predicate Function determining whether or not to include the feature
    */
   def filter(predicate: Feature[T, _, _, _] => Boolean): FeatureSpec[T] = {
-    val filteredFeatures = features
-      .filter(predicate)
-      .map[(String, Feature[T, _, _, _]), Map[String, Feature[T, _, _, _]]](f =>
-        (f.transformer.name, f))(breakOut)
-    val filteredCrossings = crossings.filter[T](filteredFeatures.contains)
+    val filteredFeatures = features.filter(predicate)
+    val featuresByName =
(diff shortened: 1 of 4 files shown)


nevillelyh:
Can you add a test?

PR #148, opened 2018-06-26
Add functions that uses Featran functions to map to a different raw feature type.

richwhitjr:
The general idea behind this review is to allow for a function that can use the internal functions from Featran to map to a new data type such as JSON or TFExample.  This becomes useful in conjunction with another #153 so that a pipeline can generate the settings file and this function in this PR can convert to TFExamples prior to saving to BigTable.

This PR is the cleanest way I could think of implementing this change and allow for flexibility.  In the future we may also want to provide a JSON converted that can be used directly in Facets Dive.  The major issue is around the TypeTag addition.  Everything works as expected but things get complex on the Java.  Currently this function is not exposed on the java side because the TypeTag is passed as null to not need to change the API.  We could also ask users to pass in the class of the function output but this would break the API and not be ideal for users.

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,6 +32,7 @@ val simulacrumVersion = "0.12.0"
 val sparkVersion = "2.3.0"
 val tensorflowVersion = "1.8.0"
 val xgBoostVersion = "0.72-20180627-1214081f"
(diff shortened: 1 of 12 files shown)


regadas:
@richwhitjr this one needs rebase 😛

PR #149, opened 2018-06-26
Default Simple Case Class Generator for FeatureSpec

richwhitjr:
This is a automatic converter for simple case classes which uses reflection to map a case class into a FeatureSpec.  In the future it would be nice to switch to a macro but this should handle many simple cases.


core/src/main/scala/com/spotify/featran/converters/CaseClassConverter.scala
```diff
@@ -0,0 +1,123 @@
...
+
+import com.spotify.featran.FeatureSpec
+import com.spotify.featran.transformers._
+
+import scala.reflect.runtime.{universe => ru}
+import scala.reflect.ClassTag
+
+object CaseClassConverter {
+  import ru._
+  private def getProperties[T: TypeTag]: Array[MethodSymbol] = {
```
regadas: 🎨 `properties` instead?

core/src/main/scala/com/spotify/featran/converters/DefaultTransform.scala
```diff
@@ -0,0 +1,31 @@
...
+ */
+
+package com.spotify.featran.converters
+
+import com.spotify.featran.transformers.{Identity, Transformer}
+
+/**
+ * Default Type Class used by the from generator for Case Class Conversions
+ */
+trait DefaultTransform[T] {
```
regadas: @richwhitjr can we use `@typeclass` where?

core/src/main/scala/com/spotify/featran/converters/DefaultTransform.scala
```diff
@@ -0,0 +1,31 @@
...
+  def apply(featureName: String): Transformer[T, _, _]
+}
+
+case object IdentityDefault extends DefaultTransform[Double] {
```
regadas: @richwhitjr here I would probably do the following:

```
object DefaultTransformers {
    implicit val identityDefault: DefaultTransform[Double] = ...
}
```

i think it gives more room to add default transforms and it's in a defined "context"

wdyt?

PR #150, opened 2018-06-26
New take on cross-building

regadas:


.travis.yml
```diff
@@ -1,25 +1,18 @@
...
-    - scala: 2.11.12
-      jdk: oraclejdk8
-    - scala: 2.12.6
-      jdk: oraclejdk8
+
+jdk: oraclejdk8
+
+env:
+  - SBT_CMD="scalafmtSbtCheck scalafmtCheck test:scalafmtCheck scalastyle test:scalastyle"
+  - SBT_CMD="coverage +test coverageReport coverageAggregate"
```
nevillelyh: The 2nd line overwrites the first?

build.sbt
```diff
@@ -156,32 +158,24 @@ lazy val core: Project = project
...
-    projectDependencies := {
-      if (scalaBinaryVersion.value == "2.12") Nil
-      else
-        Seq(
-          (projectID in core).value,
-          (projectID in tensorflow).value,
-          (projectID in xgboost).value
-        )
-    }
+    projectDependencies := Seq(
```
nevillelyh: We can probably get rid of these and rely on `.dependsOn` only?

.travis.yml
regadas: this will build a env matrix; those `SBT_CMD` will run in parallel.

build.sbt
regadas: done.

.travis.yml
```diff
@@ -1,25 +1,21 @@
...
+
+jdk: oraclejdk8
+
 matrix:
-  include:
-    - scala: 2.11.12
-      jdk: oraclejdk8
-    - scala: 2.12.6
-      jdk: oraclejdk8
+  fast_finish: true
```
regadas: this will make sure that if one of the steps fail the build will fail fast.

PR #151, opened 2018-06-27
Bump xgboost version to 0.72

regadas:


regadas:
@nevillelyh nice catch! yeah! that is actually a deal breaker!!

PR #154, opened 2018-06-29
Slight refactor of the normalization process

ravwojdyla:


tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -19,10 +19,12 @@ package com.spotify.featran
 
 import org.tensorflow.example.{Example, Features}
 import org.tensorflow.{example => tf}
+import _root_.java.util.regex.Pattern
 
 package object tensorflow {
+  private val featureNameNormalization = Pattern.compile("[^A-Za-z0-9_]")
```
richwhitjr: Patterns serializable or should this be lazy?  Although not quite clear to me at a brief glance if serialization would be attempted here.

ravwojdyla: @richwhitjr it's inside an object so we get serialization for "free".

regadas: Also Pattern is serializable.

regadas: @ravwojdyla 🎨 can we make it FeatureNameNormalization?

ravwojdyla:
@regadas why?

tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
ravwojdyla: @regadas why?

regadas: to follow scala style convention on constant names.

ravwojdyla: Does this codebase follow that rule tho, see:
 * MultiFeatureExtractor
 * FlinkCollectionType
 * MDLPDiscretizer

?

It seems like the above is the most common here.

regadas: that's why I marked the comment with 🎨 and not requested changes 😄 it's up to you

ravwojdyla: 👍

PR #155, opened 2018-06-29
Follow scala naming convention around constant naming

regadas:


nevillelyh:
Is this to make style checker happy?

regadas:
we were using multiple constant naming styles, so i think we should follow the [scala convention].(https://docs.scala-lang.org/style/naming-conventions.html#constants-values-variable-and-methods)

Implicitly it might make style checker happier 😄

PR #156, opened 2018-07-03
Convert function between Scala Objects and a Flat Keyed Type

richwhitjr:
This PR adds a new TypeClass and function that allows a user to use the functions internal to a FeatureSpec to convert to a new flat keyed feature type.  This review adds support for tf.examples but json and tablerows are other examples of possible converters.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/FlatExtractor.scala
+++ b/core/src/main/scala/com/spotify/featran/FlatExtractor.scala
@@ -56,7 +56,8 @@ object FlatExtractor {
     new FlatExtractor[M, T](setCol)
 }
 
-private[featran] class FlatExtractor[M[_]: CollectionType, T: ClassTag: FlatReader](settings: M[String])
+private[featran] class FlatExtractor[M[_]: CollectionType, T: ClassTag: FlatReader](
+  settings: M[String])
     extends Serializable {
 
   import CollectionType.ops._
--- a/core/src/main/scala/com/spotify/featran/FlatExtractor.scala
+++ b/core/src/main/scala/com/spotify/featran/FlatExtractor.scala
@@ -43,11 +43,6 @@ import scala.reflect.ClassTag
   def readStrings(name: String): T => Option[Seq[String]]
 }
 
-object FlatExtractor {
-  def apply[M[_]: CollectionType, T: ClassTag: FlatReader](setCol: M[String]): FlatExtractor[M, T] =
-    new FlatExtractor[M, T](setCol)
-}
-
 /**
  * Sometimes it is useful to store the features in an intermediate state in normally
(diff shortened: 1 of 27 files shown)


tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -101,6 +103,76 @@ package object tensorflow {
...
 
+  implicit val exampleFlatWriter: FlatWriter[Example] = new FlatWriter[tf.Example] {
+    import TensorFlowType._
+    type IF = List[NamedTFFeature]
+
+    override def writeDouble(name: String): Option[Double] => List[NamedTFFeature] =
+      (v: Option[Double]) => v.toList.map(r => NamedTFFeature(name, fromDoubles(Seq(r)).build()))
+
+    override def writeMdlRecord(name: String): Option[MDLRecord[String]] => List[NamedTFFeature] =
+      (v: Option[MDLRecord[String]]) => {
```
regadas: this will raise issues with `2.11` 😛

core/src/main/scala/com/spotify/featran/FlatExtractor.scala
```diff
@@ -43,24 +43,21 @@ import scala.reflect.ClassTag
...
- * @tparam M Collection Type
- * @tparam T The intermediate storage format for each feature.
  */
-class FlatExtractor[M[_]: CollectionType, T: ClassTag: FlatReader](settings: M[String])
+object FlatExtractor {
+  def apply[M[_]: CollectionType, T: ClassTag: FlatReader](setCol: M[String]): FlatExtractor[M, T] =
+    new FlatExtractor[M, T](setCol)
+}
+
+private[featran] class FlatExtractor[M[_]: CollectionType, T: ClassTag: FlatReader](
```
regadas: nice having it `private`

core/src/main/scala/com/spotify/featran/transformers/Binarizer.scala
```diff
@@ -52,4 +52,6 @@ private[featran] class Binarizer(name: String, val threshold: Double) extends Ma
     Map("threshold" -> threshold.toString)
 
   def flatRead[T: FlatReader]: T => Option[Any] = FlatReader[T].readDouble(name)
+  def flatWriter[T](implicit fw: FlatWriter[T]): Option[Double] => fw.IF =
```
regadas: @richwhitjr should we have it `flatWriter[T: FlatWriter]`?

richwhitjr: This signature is weird.  If you notice I am using an internally declared type as the output the returned function.  Because of that I need access to the class in the signature to write the function.

Honestly surprised this worked.

PR #159, opened 2018-07-04
Add mima to travis

regadas:


regadas:
@nevillelyh there are a lot of binary incompatibilities ... i think we should start thinking in bumping to `0.2.0`

regadas:
Also we added a bunch of new stuff.

PR #161, opened 2018-07-05
Add JSON Reader/Writer to Core

richwhitjr:
This adds a new package to core that allows for reading/writing to a JSON intermediate format.  Also because we have this typeclass implemented in core the converters can be added to transformers to make sure it works everywhere.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/json/package.scala
+++ b/core/src/main/scala/com/spotify/featran/json/package.scala
@@ -1,0 +1,152 @@
+/*
+ * Copyright 2018 Spotify AB.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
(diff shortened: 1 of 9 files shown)


core/src/test/scala/com/spotify/featran/transformers/TransformerProp.scala
```diff
@@ -101,6 +104,38 @@ abstract class TransformerProp(name: String) extends Properties(name) {
       "f3 settings" |: settings == f3.featureSettings,
       "f4 settings" |: settings == f4.featureSettings
     )
+
+    val multiSpec = MultiFeatureSpec(fsRequired)
+
+    val flatRequired = FlatExtractor.flatSpec[String, T](fsRequired)
+    val flatOptional = FlatExtractor.flatSpec[String, Option[T]](fsOptional)
```
regadas: @richwhitjr this seems to not be used?

core/src/main/scala/com/spotify/featran/json/package.scala
```diff
@@ -0,0 +1,152 @@
...
+package com.spotify.featran
+
+import com.spotify.featran.transformers.{MDLRecord, WeightedLabel}
+import io.circe._
+import io.circe.parser._
+
+/**
+ * Package for an intermediate format for feature storage using JSON
+ */
+package object json {
```
regadas: @richwhitjr I wonder if we can use the existent typeclass `JsonSerializable` to hold the serialization logic for the features. I think this way `FlatReader[String]` would just call encode and decode using the provide `JsonSerializable` typeclasses. wdyt?

regadas: @richwhitjr let's do an enhancement afterwards ... issue #162

PR #163, opened 2018-07-06
Use JsonSerializer in FlatWriter/FlatReader

regadas:
Closes #163


core/src/main/scala/com/spotify/featran/JsonSerializable.scala
```diff
@@ -20,8 +20,6 @@ package com.spotify.featran
 import io.circe.{Error, Json}
 import simulacrum.typeclass
 
-import scala.language.implicitConversions
```
regadas: Not needed since we have `-language:implicitConversions` as scalac option.

PR #164, opened 2018-07-09
Update travis build matrix

regadas:


project/build.properties
```diff
@@ -1 +1 @@
-sbt.version=1.1.6
+sbt.version=1.2.0-RC1
```
regadas: this version includes a fix around cross building for a specific version. https://github.com/sbt/sbt/issues/3698

PR #175, opened 2018-07-18
Add minor tweaks to improve inlining

regadas:


core/src/main/scala/com/spotify/featran/CrossingFeatureBuilder.scala
```diff
@@ -65,14 +65,14 @@ private class CrossingFeatureBuilder[F] private (private val fb: FeatureBuilder[
     extends FeatureBuilder[F] {
 
   private case class CrossValue(name: String, offset: Int, value: Double)
-  private var xEnabled = false // true if current transformer will be crossed
+  private[this] var xEnabled = false // true if current transformer will be crossed
```
nevillelyh: What does `[this]` scope do? So it's private within `class` & `object`?

core/src/main/scala/com/spotify/featran/CrossingFeatureBuilder.scala
```diff
@@ -56,7 +56,7 @@ private case class Crossings(map: Crossings.MAP, keys: Set[String]) {
 }
 
 object CrossingFeatureBuilder {
-  def apply[F](fb: FeatureBuilder[F], crossings: Crossings): FeatureBuilder[F] =
+  @inline def apply[F](fb: FeatureBuilder[F], crossings: Crossings): FeatureBuilder[F] =
```
nevillelyh: This is for performance reasons? How noticeable is it?

core/src/main/scala/com/spotify/featran/CrossingFeatureBuilder.scala
regadas: yes and also makes the compiler not generate accessor methods, do this avoids extra calss.

core/src/main/scala/com/spotify/featran/CrossingFeatureBuilder.scala
regadas: hard to say, i think it is the typical YMMV but i think it's good to hint the compiler to try to inline some of the things as much as possible to avoid extra calls.

regadas:
I probably missed some ... this are a the ones that really caught my eye.

PR #176, opened 2018-07-18
Small tensorflow cleanup

regadas:


tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -17,20 +17,26 @@
...
-case class NamedTFFeature(name: String, f: tf.Feature)
-
 package object tensorflow {
-  private val FeatureNameNormalization = Pattern.compile("[^A-Za-z0-9_]")
+
+  private[this] object FeatureNameNormalization {
+    private[this] val NamePattern = Pattern.compile("[^A-Za-z0-9_]")
+
+    val normalize: String => String = fn =>
+      NamePattern.matcher(fn).replaceAll("_")
```
richwhitjr: Do we want to add memoization here also?

tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -29,7 +31,13 @@ package object tensorflow {
   private[this] object FeatureNameNormalization {
     private[this] val NamePattern = Pattern.compile("[^A-Za-z0-9_]")
 
-    val normalize: String => String = fn => NamePattern.matcher(fn).replaceAll("_")
+    val normalize: String => String = {
```
regadas: @nevillelyh what do you think of this?

tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
regadas: @richwhitjr yup I guess it can improve things a little bit

PR #178, opened 2018-07-18
Optimize FlatReader read

### LINKED ISSUE #174: FlatExtractor performance.
Currently FlatExtractor parses the entire flat object newly for each flat record.  Instead we should parse the flat record once and index into the parsed record to get each feature.  There are two ways we can go about doing this.

1. The typeclass can hold state on the open object to save on the parsing.
2. Follow the Writer path and add a read method that takes in a flat record and outputs a map of features.

regadas:
Closes #174


regadas:
@richwhitjr PTAL

PR #181, opened 2018-07-25
extract with subset settings java bindings

Rafff:
This PR aims to add Java bindings to the library, in order to access extractWithSubsetSettings functions from Java.


alaiacano:
I'm going to close this one and re-open on my branch with tests, since @Rafff is on vacation :)

PR #182, opened 2018-07-30
Laiacano/java extract subsettings

alaiacano:
This PR aims to add Java bindings to the library, in order to access extractWithSubsetSettings functions from Java.


regadas:
@alaiacano tests failing,

PR #235, opened 2019-02-19
Add indicator feature

derenrich:
Add support for indicator features (1 if a feature is present / 0 otherwise)

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/Indicator.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/Indicator.scala
@@ -17,9 +17,8 @@
 
 package com.spotify.featran.transformers
 
-import com.spotify.featran.{FlatReader, FlatWriter, FeatureBuilder}
+import com.spotify.featran.{FeatureBuilder, FlatReader, FlatWriter}
 import com.twitter.algebird.Aggregator
-
 
 /**
  * Transform an optional 1D feature to an indicator variable indicating presence.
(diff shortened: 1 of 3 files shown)


regadas:
Yeah I’ll have a look

core/src/main/scala/com/spotify/featran/transformers/Indicator.scala
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2017 Spotify AB.
```
regadas: can you update to 2019?

core/src/test/scala/com/spotify/featran/Fixtures.scala
```diff
@@ -88,6 +88,7 @@ object Fixtures {
     .required(_.s3)(HashNHotWeightedEncoder("hash-n-hot-weighted"))
     .required(_.s1)(HeavyHitters("heavy-hitters", 10, 0.001, 0.001, 1))
     .required(_.x)(Identity("id"))
+    .optional(_.xo)(Indicator("indic"))
```
regadas: why not have the full name?

core/src/test/scala/com/spotify/featran/transformers/IndicatorSpec.scala
```diff
@@ -0,0 +1,29 @@
+/*
+ * Copyright 2017 Spotify AB.
```
regadas: 🎨 2019

core/src/main/scala/com/spotify/featran/transformers/Indicator.scala
```diff
@@ -0,0 +1,63 @@
...
+
+  /**
+   * Create a new [[Indicator]] from a settings object
+   * @param setting Settings object
+   */
+  def fromSettings(setting: Settings): Transformer[Double, Unit, Unit] =
+    Indicator(setting.name)
+}
+
+private[featran] class Indicator(name: String) extends OneDimensional[Double, Unit, Unit](name) {
```
regadas: you could potentially just use `MapOne` and override the map function

```scala
class Indicator(name: String) extends MapOne[Double](name) {
  override def map(a: Double): Double = 1

  override def flatRead[T: FlatReader]: T => Option[Any] = FlatReader[T].readDouble(name)

  override def flatWriter[T](implicit fw: FlatWriter[T]): Option[Double] => fw.IF =
    fw.writeDouble(name)
}
```

derenrich: this is ... a good point

derenrich:
comments addressed

PR #288, opened 2019-09-16
decouple XGBoost, update cross build

nevillelyh:


Helw150:
@nevillelyh Do you have plans to push a new release to the Maven Repository with these cross-compilation changes? I'd love to use Featran with Scala 2.12 and XGBoost (which will support 2.12 as it's primary target with it's next release https://github.com/dmlc/xgboost/pull/4574)

nevillelyh:
Will do. Sonatype is uber-slow the last few days so it might take a while.

Helw150:
Thank you!

Issue #103, opened 2018-04-17
Featran is hard to extend with custom transformers


yonromai:
Featran users complained that  “library scala skillz” are required for them to implement the custom transformers they need.

fallonchen:
@yonromai Do you think it would it be sufficient to add a guide to writing transformers? I'd estimate the Scala skill level needed for this is is intermediate, not advanced, if a good explanation is provided.

yonromai:
Yes I think we can start with a guide and share it with some internal customers to see if that's enough. Definitely a good start.

nevillelyh:
No activity in 1+ yr and doesn't seem a huge issue, closing.

PR #311, opened 2019-10-26
Add contramap method to Transformer

JD557:
Adds a `contramap` method to the `Transformer`, to make it slightly easier to write custom Transformers (which seems to be a pain, right now: #103).

This could already mostly be done in the `require` lambda, but having `contramap` on the `Transformer` makes it easier to reuse custom transformers.

Example usage:
```scala
import com.spotify.featran._
import com.spotify.featran.transformers._

case class Student(name: String, age: Int, height: Double)
case class Course(students: List[Student])
val data = Seq(
  Course(
    List(
      Student("Anna", 20, 1.65),
      Student("Brad", 23, 1.72),
      Student("Peter", 19, 1.63)
    )
  ),
  Course(
    List(
      Student("Alice", 30, 1.59),
      Student("Bob", 32, 1.67),
      Student("Chloe", 25, 1.55),
      Student("Dennis", 27, 1.80)
    )
  )
)

// Using base Transformers to create more complex Transformers
def AvgMinMaxScaler(name: String) =
  MinMaxScaler(name).contramap((x: List[Double]) => x.sum / x.size)
def SizeMinMaxScaler(name: String) =
  MinMaxScaler(name).contramap((x: List[_]) => x.size)
// Any function can be converted to a transformer by contramapping the Identity Transformer
def CustomTransformer(name: String) =
  Identity(name).contramap { (x: Course) =>
    if (x.students.size % 2 == 0) 3.14
    else x.students.head.age + x.students.head.height
  }

val fs = FeatureSpec.of[Course]
  .required(_.students)(SizeMinMaxScaler("class-size-min-max"))
  .required(_.students.map(_.age.toDouble))(AvgMinMaxScaler("student-age-avg-min-max"))
  .required(_.students.map(_.height))(AvgMinMaxScaler("student-height-avg-min-max"))
  .required(identity[Course])(CustomTransformer("custom"))

val fe = fs.extract(data)
val names = fe.featureNames
val features = fe.featureValues[Seq[Double]]
```

I think there's still a lot of stuff to be done to make `Transformer`s easier to create (it would be awesome to have other operations like `map`, `zip`,...), but that might require a large refactor.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
+++ b/core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
@@ -37,7 +37,7 @@ trait SettingsBuilder {
  * @tparam B aggregator intermediate type
  * @tparam C aggregator summary type
  */
-abstract class Transformer[-A, B, C](val name: String) extends Serializable {
+abstract class Transformer[-A, B, C](val name: String) extends Serializable { self =>
 
   require(name != null && name.nonEmpty, "name cannot be null or empty")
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
```diff
@@ -70,6 +70,31 @@ abstract class Transformer[-A, B, C](val name: String) extends Serializable {
...
+        override def semigroup = self.aggregator.semigroup
+        override def present(b: B): C = self.aggregator.present(b)
+      }
+      def buildFeatures(a: Option[AA], c: C, fb: FeatureBuilder[_]): Unit =
+        self.buildFeatures(a.map(f), c, fb)
+      def decodeAggregator(s: String): C = self.decodeAggregator(s)
+      def encodeAggregator(c: C): String = self.encodeAggregator(c)
+      def featureDimension(c: C): Int = self.featureDimension(c)
+      def featureNames(c: C): Seq[String] = self.featureNames(c)
+      def flatRead[T](implicit evidence$1: FlatReader[T]): T => Option[Any] =
```
regadas: 🎨 we can just use `FlatReader` as a context bound

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
```diff
@@ -70,6 +70,31 @@ abstract class Transformer[-A, B, C](val name: String) extends Serializable {
...
   protected def names(n: Int): Stream[String] = (0 until n).toStream.map(nameAt)
 
+  /**
+   * Builds a new transformer with an extra input preprocessing step
+   * @param f input preprocessing function
+   */
+  def contramap[AA](f: AA => A): Transformer[AA, B, C] = {
+    val self = this
+    new Transformer[AA, B, C](name) {
+      val aggregator: Aggregator[AA, B, C] = new Aggregator[AA, B, C] {
```
regadas: 🎨 Add missing `override`

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
JD557: Addressed in 014c640

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
JD557: Addressed in 014c640

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
```diff
@@ -70,6 +70,31 @@ abstract class Transformer[-A, B, C](val name: String) extends Serializable {
   protected def nameAt(n: Int): String = name + '_' + n
   protected def names(n: Int): Stream[String] = (0 until n).toStream.map(nameAt)
 
+  /**
+   * Builds a new transformer with an extra input preprocessing step
+   * @param f input preprocessing function
+   */
+  def contramap[AA](f: AA => A): Transformer[AA, B, C] = {
+    val self = this
```
nevillelyh:
```suggestion
```

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
```diff
@@ -70,6 +70,31 @@ abstract class Transformer[-A, B, C](val name: String) extends Serializable {
   protected def nameAt(n: Int): String = name + '_' + n
   protected def names(n: Int): Stream[String] = (0 until n).toStream.map(nameAt)
 
+  /**
+   * Builds a new transformer with an extra input preprocessing step
+   * @param f input preprocessing function
+   */
+  def contramap[AA](f: AA => A): Transformer[AA, B, C] = {
```
nevillelyh:
```suggestion
  def contramap[AA](f: AA => A): Transformer[AA, B, C] = { self =>
```

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
JD557: Addressed in 6fbfbc0

core/src/main/scala/com/spotify/featran/transformers/Transformer.scala
JD557: Addressed in 6fbfbc0

This suggestion doesn't actually compile, so I had to declare `self` on the class declaration, I hope that's OK:

```scala
abstract class Transformer[-A, B, C](val name: String) extends Serializable { self =>
```

nevillelyh: Right, I misread :joy:

PR #319, opened 2019-11-25
clean up typos, unused & other warnings

nevillelyh:


core/src/main/scala/com/spotify/featran/json/Implicits.scala
```diff
@@ -17,7 +17,6 @@
 
 package com.spotify.featran.json
 
-import cats.syntax.either._
```
regadas: this one is need for 2.11

PR #320, opened 2019-11-26
add Scala 2.13 support

nevillelyh:


core/src/main/scala/com/spotify/featran/CanBuild.scala
```diff
@@ -0,0 +1,67 @@
...
+
+// Workaround for CanBuildFrom not serializable
+trait CanBuild[T, M[_]] extends Serializable {
+  def apply(): mutable.Builder[T, M[T]]
+}
+
+object CanBuild {
+  // Collection types in _root_.scala.*
+  //scalastyle:off public.methods.have.type
+  implicit def iterableCB[T] = new CanBuild[T, Iterable] {
```
regadas: would it be ok to add some return types here?

build.sbt
```diff
@@ -61,23 +76,11 @@ lazy val publishSettings = Seq(
     username <- sys.env.get("SONATYPE_USERNAME")
     password <- sys.env.get("SONATYPE_PASSWORD")
   } yield Credentials("Sonatype Nexus Repository Manager", "oss.sonatype.org", username, password)).toSeq,
-  publishTo := sonatypePublishToBundle.value,
+  publishTo := Some(
+    if (isSnapshot.value) Opts.resolver.sonatypeSnapshots else Opts.resolver.sonatypeStaging
+  ),
   releaseCrossBuild := true,
   releasePublishArtifactsAction := PgpKeys.publishSigned.value,
-  releaseProcess := Seq[ReleaseStep](
```
regadas: Are you still seeing issues with this? I would like this to return; I tried it out and it seems to work as expected.

core/src/main/scala/com/spotify/featran/FeatureBuilder.scala
```diff
@@ -180,60 +179,9 @@ case class NamedSparseArray[@specialized(Float, Double) T](
...
-    def apply(): mutable.Builder[T, M]
-  }
-
-  private def newCB[T, M](f: () => mutable.Builder[T, M]) = new CanBuild[T, M] {
-    override def apply(): mutable.Builder[T, M] = f()
-  }
-
-  // Collection types in _root_.scala.*
-  //scalastyle:off public.methods.have.type
-  implicit def traversableCB[T] = newCB(() => Traversable.newBuilder[T])
```
regadas: we could keep these and ref the new ones under `CanBuild` implicit scope. However we can always just remove them and bum,p version to `0.5.0`

regadas: MiMa will be mainly complain about this ... if we bump version I don't care that much about bin compat

nevillelyh: Yeah plus people really shouldn't be using `Traversable` directly. This was supposed to be a catch all for `predef` collection types.

build.sbt
nevillelyh: At least `release skip-tests` doesn't work with this, and there's no way of just publishing to staging without actually releasing the artifacts for real, so there's no way to dry-run this? Not going to waste more time on this anymore.

regadas:
```
  releaseProcess := Seq[ReleaseStep](
    checkSnapshotDependencies,
    inquireVersions,
    runClean,
    ReleaseStep { st: State =>
      if (!st.get(ReleaseKeys.skipTests).getOrElse(false)) {
        releaseStepCommandAndRemaining("+test")(st)
      } else {
        st
      }
    },
    setReleaseVersion,
    commitReleaseVersion,
    tagRelease,
    releaseStepCommandAndRemaining("+publishSigned"),
    releaseStepCommand("sonatypeBundleRelease"),
    setNextVersion,
    commitNextVersion,
    pushChanges
  ),
```

PR #426, opened 2020-10-29
Remove simulacrum

regadas:
One step closer to Dotty support

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -37,6 +37,8 @@ val shapelessDatatypeVersion = "0.2.0"
 
 val previousVersion = "0.7.0"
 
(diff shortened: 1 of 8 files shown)


regadas (commented):
@nevillelyh currently this is not really a removable. I'm just replacing the macro with generated code. The reason is that the `ops` provided are handy and I'm not sure if their's ppl out there taking advantage of them.

Do we go ahead with this and deprecate on later stage or remove them completely now?

regadas:
@nevillelyh went ahead and merged this as it was ageing a bit ... let me know if you think it's better to have this removed and we can release a new backwards incompatible version.

PR #464, opened 2021-01-18
Update xgboost to 1.3.1

### LINKED ISSUE #467: Can't mix `featran-xgboost` dependency with newer versions of xgboost
```
java.lang.NoSuchMethodError: 'int ml.dmlc.xgboost4j.LabeledPoint.size()'
        at ml.dmlc.xgboost4j.java.DataBatch$BatchIterator.next(DataBatch.java:43)
        at ml.dmlc.xgboost4j.java.XGBoostJNI.XGDMatrixCreateFromDataIter(Native Method)
        at ml.dmlc.xgboost4j.java.DMatrix.<init>(DMatrix.java:54)
```

regadas:
Fixes #467

### DIFF
--- a/build.sbt
+++ b/build.sbt
@@ -32,7 +32,7 @@ val scioVersion = "0.10.0-beta2"
 val simulacrumVersion = "1.0.1"
 val sparkVersion = "3.1.0"
 val tensorflowVersion = "1.15.0"
-val xgBoostVersion = "0.90"
+val xgBoostVersion = "1.3.1"
 val shapelessDatatypeVersion = "0.2.0"
 
 val previousVersion = "0.7.0"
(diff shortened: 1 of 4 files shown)


build.sbt
```diff
@@ -32,7 +32,7 @@ val scioVersion = "0.10.0-beta2"
 val simulacrumVersion = "1.0.1"
 val sparkVersion = "3.1.0"
 val tensorflowVersion = "1.15.0"
-val xgBoostVersion = "0.90"
+val xgBoostVersion = "1.3.1"
```
regadas, line 35: This is not really used anywhere and left it here just that we know which version are we targeting with the copied `LabeledPoint`

regadas:
we will need to release 0.8.0 since this will not be been compat

nevillelyh:
Is that why mima failed? Should we merge & fix that after?

regadas:
@nevillelyh yea that's right! we can merge now

PR #469, opened 2021-01-20
Update TensorFlow

regadas:
* Removed shapeless dependency 
* Compiles against 3.0.0-M3


tensorflow/src/main/scala/com/spotify/featran/tensorflow/package.scala
```diff
@@ -70,11 +70,8 @@ package object tensorflow {
     override def newBuilder: FeatureBuilder[tf.Example] = TensorFlowFeatureBuilder()
   }
 
-  /** [[FeatureBuilder]] for output as TensorFlow `Example` type. */
-  implicit def tensorFlowFeatureBuilder: FeatureBuilder[tf.Example] = TensorFlowFeatureBuilder()
-
-  implicit val exampleFlatReader: FlatReader[tf.Example] = new FlatReader[tf.Example] {
-    import TensorFlowType._
```
nevillelyh, line 77: So we were actually not using the shapeless based `TensorFlowType[T]` converters, just implicits/utilities from that package?

regadas: yea, basically just utility methods from that package.

Issue #51, opened 2017-10-11
Add Segment Indices transformer


andrewsmartin:
See https://github.com/tensorflow/transform/blob/master/tensorflow_transform/mappers.py#L374

nevillelyh:
This should be a simple map-only transformer over an `Array[Int]`.

PR #536, opened 2021-06-20
Add Segment Indices transformer

aaronquantexa:
Had a play with solving this one: https://github.com/spotify/featran/issues/51

CL: 

- Implements an equivalent functional output to: https://github.com/tensorflow/transform/blob/master/tensorflow_transform/mappers.py#L1209  using the Transformer API on `Array[Int]` . Is efficient because it iterates over the input Array once (`O(n)`) using a fold.
-  Add support for `Int` data types with Feature Builder specifically for Segment Indices
-  Add example usage of new Transformer in Examples in plain Scala

Testing-wise, I've verified the results on a few arrays manually by running the Example as well as adding a property test which checks the result on 100 randomly generated (and valid) arrays. It does not 100% follow the pattern of other property tests in the repo which check for feature column names etc.

### DIFF
--- a/core/src/main/scala/com/spotify/featran/CanBuild.scala
+++ b/core/src/main/scala/com/spotify/featran/CanBuild.scala
@@ -63,11 +63,7 @@ object CanBuild {
     override def apply(): mutable.Builder[T, Array[T]] = Array.newBuilder[T]
   }
 
-  implicit def IntSeqCB: CanBuild[Int, Seq] = new CanBuild[Int, Seq] {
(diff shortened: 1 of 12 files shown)


aaronquantexa:
In the file: _SegmentIndicesSpec.scala,_ I had a functional-looking version of implementing the Arbitrary `randomIncreasingArray` which used a fold but it was less clear than the mutable alternative, which suits the dynamic style of the logic, so went with that implementation.

PR #677, opened 2023-02-10
Add scala3 support for the java module

RustedBones:
`java` module wasn't building with scala 3 due to type conflict as explained in https://github.com/lampepfl/dotty/discussions/16797#discussioncomment-4845603.
Proposed solution to change compiling order worked

### DIFF
--- a/java/src/main/java/com/spotify/featran/java/SerializableFunction.java
+++ b/java/src/main/java/com/spotify/featran/java/SerializableFunction.java
@@ -18,7 +18,6 @@
 package com.spotify.featran.java;
 
 import java.io.Serializable;
+import java.util.function.Function;
 
-public interface SerializableFunction<InputT, OutputT> extends Serializable {
-  OutputT apply(InputT input);
-}
+public interface SerializableFunction<InputT, OutputT> extends Function<InputT, OutputT>, Serializable {}
--- a/java/src/main/scala/com/spotify/featran/java/JavaOps.scala
+++ b/java/src/main/scala/com/spotify/featran/java/JavaOps.scala
@@ -20,6 +20,7 @@ package com.spotify.featran.java
 import java.lang.{Double => JDouble}
 import java.util.function.BiFunction
 import java.util.{Collections, List => JList, Optional}
+import java.util.function.{Function => JFunction}
 
 import com.spotify.featran._
 import com.spotify.featran.tensorflow._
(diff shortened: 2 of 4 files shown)


java/src/main/java/com/spotify/featran/java/SerializableFunction.java
```diff
@@ -18,7 +18,6 @@
 package com.spotify.featran.java;
 
 import java.io.Serializable;
+import java.util.function.Function;
 
-public interface SerializableFunction<InputT, OutputT> extends Serializable {
-  OutputT apply(InputT input);
-}
+public interface SerializableFunction<InputT, OutputT> extends Function<InputT, OutputT>, Serializable {}
```
regadas, line 23: I think this will break bin compat. Is this change really necessary? 🤔

java/src/main/scala/com/spotify/featran/java/JavaOps.scala
```diff
@@ -20,25 +20,26 @@ package com.spotify.featran.java
...
-  def optionalFn[I, O](f: SerializableFunction[I, Optional[O]]): I => Option[O] =
-    (input: I) => {
-      val o = f(input)
-      if (o.isPresent) Some(o.get()) else None
-    }
+  def requiredFn[I, O](f: JFunction[I, O]): I => O =
+    f.asScala
+
+  def optionalFn[I, O](f: JFunction[I, Optional[O]]): I => Option[O] =
+    f.asScala.andThen(_.asScala)
```
RustedBones, line 42: Binary compat is broken here.
Since I change the compile order, it was not possible to reference the project's java code in scala.

This can cause issue if pple use a previous `SerializableFunction` wich is not extending `Function`

regadas: Ah yup this is also a clear break; I would like to keep the `SerializableFunction` type here.

RustedBones:
Publishing was aborted due to
```
[error]  * static method optionalFn(com.spotify.featran.java.SerializableFunction)scala.Function1 in class com.spotify.featran.java.JavaOps's type is different in current version, where it is (java.util.function.Function)scala.Function1 instead of (com.spotify.featran.java.SerializableFunction)scala.Function1
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.spotify.featran.java.JavaOps.optionalFn")
[error]  * static method requiredFn(com.spotify.featran.java.SerializableFunction)scala.Function1 in class com.spotify.featran.java.JavaOps's type is different in current version, where it is (java.util.function.Function)scala.Function1 instead of (com.spotify.featran.java.SerializableFunction)scala.Function1
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.spotify.featran.java.JavaOps.requiredFn")
[error]  * method requiredFn(com.spotify.featran.java.SerializableFunction)scala.Function1 in object com.spotify.featran.java.JavaOps's type is different in current version, where it is (java.util.function.Function)scala.Function1 instead of (com.spotify.featran.java.SerializableFunction)scala.Function1
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.spotify.featran.java.JavaOps.requiredFn")
[error]  * method optionalFn(com.spotify.featran.java.SerializableFunction)scala.Function1 in object com.spotify.featran.java.JavaOps's type is different in current version, where it is (java.util.function.Function)scala.Function1 instead of (com.spotify.featran.java.SerializableFunction)scala.Function1
[error]    filter with: ProblemFilters.exclude[IncompatibleMethTypeProblem]("com.spotify.featran.java.JavaOps.optionalFn")
[error] java.lang.RuntimeException: Failed binary compatibility check against com.spotify:featran-java_2.12:0.8.0 (e:info.apiURL=https://oss.sonatype.org/service/local/repositories/snapshots/archive/com/spotify/featran-docs_2.12/0.8.0-5-ee94d16-SNAPSHOT/featran-docs_2.12-0.8.0-5-ee94d16-SNAPSHOT-javadoc.jar/!/index.html, e:info.versionScheme=early-semver)! Found 4 potential problems (filtered 1)
```

java/src/main/scala/com/spotify/featran/java/JavaOps.scala
regadas: Maybe we shoud just implement `SerializableFunction` as a `trait`

PR #679, opened 2023-02-12
Binary compatible scala 3 java module - Alternative

RustedBones:
@regadas alternative changes to support scala 3 in the `java` module


regadas (commented):
I like this since we are not messing with the compiler order. Though I think we are still prone to the same forward compat issues.

Tbh, I'm leaning towards just breaking and making it nicer than make it clunky. That said I think we could go ahead and just change the signatures of `extract` and `extracWithSettings`

regadas (approved):

